PR #44, opened 2016-04-20
Introduce Storage driver

rgrover:
I"ve split the header https://github.com/ARMmbed/flash-abstraction/blob/master/flash-abstraction/Driver_Storage.h into two files as expected by CMSIS. I've also added sample code to demonstrate the use of Driver_Storage. Look forward to your feedback.

@RobertRostohar @ReinhardKeil

CC: @meriac


rgrover:
duplicated by #55

PR #51, opened 2016-04-26
Removed GCC compiler warning.

GuentherMartin:


ilg-ul:
```
(pFunc)((uint32_t)&__StackTop)
```

constructs like this may silence the warning, but generally it is not ok to cast a pointer to another pointer via an integer.

PR #56, opened 2016-05-09
Updated NVIC functions.

GuentherMartin:


ilg-ul:
I noticed that in the doxygen file you changed the definitions of some return values from:

```
\return             0  if interrupt is assigned to Secure
\return             1  if interrupt is assigned to Non Secure
```

to

```
\returns            
        - 0  if interrupt is assigned to Secure
        - 1  if interrupt is assigned to Non Secure
```

I did not check if this syntax looks better, but, as far as I know, the [official syntax](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdretval) for multiple return values is:

```
\retval             0  if interrupt is assigned to Secure
\retval             1  if interrupt is assigned to Non Secure
```

PR #77, opened 2016-08-28
Include Make-based documentation generation

senselessDev:
Besides the Batch generation of the documentation, I think it would be nice to be able to use make instead on UNIX systems.

### DIFF
--- a/CMSIS/DoxyGen/Core/Makefile
+++ b/CMSIS/DoxyGen/Core/Makefile
@@ -1,0 +1,4 @@
+doc:
+	doxygen core.dxy
+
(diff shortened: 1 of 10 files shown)


ilg-ul:
good luck convincing the Keil team to use anything else than the magnificent Windows.

JonatanAntoni:
Unfortunately we do now have capacity here to really maintain this enhancement. I appreciate the need for a non-batch script for generating the documentation on non-windows machines. 
@senselessDev would you like to maintain this enhancement on your own for the moment? As we move further we might catch this up again.

PR #136, opened 2017-01-17
Fix M4 targets without FPU

bulislaw:
Some platforms with M4 don't have FPU eg FRDM-KW24D

CC: @RobertRostohar

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/ARM/irq_cm4f.s
+++ b/CMSIS/RTOS2/RTX/Source/ARM/irq_cm4f.s
@@ -74,8 +74,10 @@ SVC_Context
 
 SVC_ContextSave
                 STMDB    R12!,{R4-R11}          ; Save R4..R11
+#ifdef __FPU_PRESENT
                 TST      LR,#0x10               ; Check if extended stack frame
                 VSTMDBEQ R12!,{S16-S31}         ;  Save VFP S16.S31
+#endif
 
                 STR      R12,[R1,#TCB_SP_OFS]   ; Store SP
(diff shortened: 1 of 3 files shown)


RobertRostohar:
M4 without FPU should use the same IRQ Handler file as M3 (irq_cm3.s). This irq_cm4f.s is only for M4 with FPU.

Modifications from this PR are therefore not needed.

RobertRostohar:
You can take a look the ARM/GCC/IAR projects for building the library or the Examples to see which files needs to be included.

RTX5 has been tested on Cortex-M0, Cortex-M3, Cortex-M4/7 with and without FPU and Cortex M23/M33.

bulislaw:
Right, I'm porting the top of the develop branch to mbed and trying to flush all the changes to CMSIS, that are required and we haven't talked before. So the mbed's build system uses TARGET_<CORE> directories, and every M4 platform gets the irq_cm4.s file. So we can't really adapt it for 'no-fpu' M4, we can carry on this patch if you think it's not something you want to merge.

RobertRostohar:
This PR would break non-mbed environment for M4 with FPU because `__FPU_PRESENT `is not defined by the compilers (it is a CMSIS core define).

If FPU is present or not needs to be extracted from compilers built-in symbols which can be tricky.

Therefore I would not merge this but rather leave it for mbed to patch.

Note that you should also exclude the code that handles the FPU Lazy State. Make a file compare between irq_cm3.s and irq_cm4f.s.

bulislaw:
Ok, thanks.

bulislaw:
I'll close this PR then.

PR #139, opened 2017-01-31
Improve NVIC virtualization mechanism

Patater:
To make the use of the CMSIS_NVIC_VIRTUAL feature easier, include a header file with a standard name. This header file will be provided by the client of the CMSIS_NVIC_VIRTUAL feature to define which implementations of the NVIC functions should be used.

Also, additionally allow the virtualization NVIC_SystemReset.

@RobertRostohar


GuentherMartin:
Hello Patater,
do these changes also apply to the other core include files?

Patater:
Thanks for the review and merge.

We should enable the same sort of virtualization ability on Cortex-M7, but the virtualization feature is not yet used on Cortex-M7.

For all other platforms, I don't yet foresee the need for the CMSIS_NVIC_VIRTUAL feature.

PR #143, opened 2017-02-09
Bugfix: terminate_list may be corrupted upon join() or detach()

DanielBoehmer:
terminate_list is supposed to be a double linked list. At least osRtxThreadListUnlink() expects one. The thread->prev pointer of already terminated threads were kept to nullptr, building a sinle linked list. The function osRtxThreadListUnlink() may corrupt the terminate_list when join() or detach() is called in a different order than the threads that terminated before.


RobertRostohar:
Thanks for reporting this bug. PR has been merged and updated (with NULL pointer check).

Issue #149, opened 2017-02-16
IAR support for RTX (can't compile sources)


0xc0170:
Include is here : https://github.com/ARM-software/CMSIS_5/blob/c5e87e54ffe69be548823be030b0d2169eb2b707/CMSIS/Core/Include/cmsis_compiler.h#L66

But I can not locate it in the tree, thus some definitions for IAR are missing.  The lines referenced above do not look good neither, as for IAR there are lot of macros that should probably live in the iar specific header file (same is done for gcc, and other toolchains).

jkrech:
The file cmsis_iar.h is shipped by IAR as part of their tool-chain.
If you have EWARM installed you should be able to locate it there or if you are not using IAR your tool-chain should ignore this section of the file completely as you must not have __ICCARM__ being defined.

0xc0170:
Thanks ! I was able to find the file, however I am still hitting the same problem that I suspected due to missing header file. I explored that cmsis_iar.h does not contain any definitions for this error:

```
  #error "Unknown ARM Architecture!"
   ^
"rtos\rtx2\TARGET_CORTEX_M\core_cm.h",56  Fatal error[Pe035]: #error directive: "Unknown ARM Architecture!"
```

GCC defines them, for armcc there is a translation between their symbols and this __ARM_ARCH. What about for IAR? I noticed that this __ARM_ARCH macro is defined in one of .ewp files that does not look right. I am not using this .ewp but only sources from this repository.

0xc0170:
Should this be part of cmsis_compiler.h for IAR specific part:

```
#if (__CORE__ == __ARM6M__) || (__CORE__ == __ARM6SM__)
#define __ARM_ARCH_6M__ 1
#elif (__CORE__ == __ARM7M__)
#define __ARM_ARCH_7M__ 1
#elif (__CORE__ == __ARM7EM__)
#define __ARM_ARCH_7EM__ 1
#endif
```

Then those symbols __ARM_ARCH_xx__ could be removed from .ewp project files, the sources should compile.

I can't find what version of IAR was used for those project files, I got 7.70.1 and receive a warning when I open a project, that means it was generated by older version of IAR (broken options detected warning).

0xc0170:
Once I defined __ARM_ARCH_ macros for IAR as above, I progressed with errors in our code base, and hit this one:

```
rtx_msgqueue.c",111  Error[Ta117]: IT-block ends prematurely, or there is a label within the IT-block
[...]

PR #150, opened 2017-02-16
cmsis: fix IAR architecture macros

0xc0170:
``__CORE__`` is IAR specific, similar to ``__ARM_ARCH_`` macros that GCC defines. Plus I removed those macros from ewp file.

reference: #149 

Note: I am working on different code base where I was testing this patch, please test it or let me know how I could test this within this code base (run any local tests). I tried to open .eww project but as in the reference, I probably have newer IAR version thus the project is not compatible.

### DIFF
--- a/CMSIS/Core/Include/cmsis_compiler.h
+++ b/CMSIS/Core/Include/cmsis_compiler.h
@@ -53,6 +53,7 @@
  */
 #elif defined ( __ICCARM__ )
 
+
   #ifndef   __ASM
     #define __ASM                                  __asm
   #endif
@@ -64,6 +65,21 @@
   #endif
 
   #include <cmsis_iar.h>
+
+  /* CMSIS compiler control architecture macros */
+  #if (__CORE__ == __ARM6M__) || (__CORE__ == __ARM6SM__)
+    #ifndef __ARM_ARCH_6M__
+      #define __ARM_ARCH_6M__                      1
+    #endif
+  #elif (__CORE__ == __ARM7M__)
+    #ifndef __ARM_ARCH_7M__
+      #define __ARM_ARCH_7M__                      1
+    #endif
+  #elif (__CORE__ == __ARM7EM__)
+    #ifndef __ARM_ARCH_7EM__
+      #define __ARM_ARCH_7EM__                     1
+    #endif
+  #endif
 
   #ifndef   __NO_RETURN
     #define __NO_RETURN                            __noreturn
--- a/CMSIS/RTOS2/RTX/Library/IAR/IDE/RTX_CM.ewp
+++ b/CMSIS/RTOS2/RTX/Library/IAR/IDE/RTX_CM.ewp
@@ -182,7 +182,7 @@
         <option>
           <name>CCDefines</name>
           <state>NDEBUG</state>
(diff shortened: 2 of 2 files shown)


JonatanAntoni:
Looks good and functional.

CMSIS/Core/Include/cmsis_compiler.h
```diff
@@ -53,6 +53,15 @@
  */
 #elif defined ( __ICCARM__ )
 
+  /* CMSIS compiler control architecture macros */
```
JonatanAntoni: Anders Lundgren from IAR aims to include these defines into their `cmsis_iar.h`. He announce the release 8.10 for today which should contain the attached version of their header file. Please review the changes shipped by IAR. If the changes are working we can skip the defines here. Please update the PR accordingly.

0xc0170: It's not part of IAR 7.8 that we use, neither earlier versions. I can edit this to check if its not already defined, then it would be valid for any IAR version.

0xc0170: I updated it now we dont redefine if already defined. Or we can use similar what they use: ``__ARM_ARCH`` ?

CMSIS/Core/Include/cmsis_compiler.h
```diff
@@ -53,6 +53,21 @@
...
+  #if (__CORE__ == __ARM6M__) || (__CORE__ == __ARM6SM__)
```
JonatanAntoni: We should do some further clean up here. I suggest to:
- [ ] @JonatanAntoni check with IAR whether they plan to port this fix back to older release branches.
- [ ] @0xc0170 move the define block behind the include to cmsis_iar.h
- [ ] @0xc0170 check whether it is possible to exclude this block for compiler versions 8.10 and later. this would highlight the backward compatibility reason of those defines.

0xc0170: OK let me know once we get IAR statement. I can do the second and not fully certain about the third point.

> @0xc0170 check whether it is possible to exclude this block for compiler versions 8.10 and later. this would highlight the backward compatibility reason of those defines.

We could do it backward-forward compatible. If I move that block as you suggested below cmsis_iar inclusion, then we just check if not yet defined those required macros, and we define them. This should be backward compatible, and forward compatible.

> @0xc0170 check whether it is possible to exclude this block for compiler versions 8.10 and later. this would highlight the backward compatibility reason of those defines.

I can add a comment to this new addition that newer version of IAR (8.x) already define these macros, but older version dont. To explain the reasoning

JonatanAntoni: Hi @0xc0170. Are the __CORE__ issues solved in the meanwhile by using latest IAR patch level? Please update this PR to a bare minimum.

0xc0170: Will report, downloading latest 7.80.4 and can test it without these patch

0xc0170: 7.80.4 - ``[Fatal Error] core_cm.h@57,0: [Pe035]: #error directive: "Unknown ARM Architecture!"`` without this patch

PR #159, opened 2017-03-02
RTX: Add headers to doxygen groups

bulislaw:
We talked about it during one of our calls.

CC: @RobertRostohar

### DIFF
--- a/CMSIS/RTOS/RTX/INC/RTX_CM_lib.h
+++ b/CMSIS/RTOS/RTX/INC/RTX_CM_lib.h
@@ -32,6 +32,9 @@
  * POSSIBILITY OF SUCH DAMAGE.
  *---------------------------------------------------------------------------*/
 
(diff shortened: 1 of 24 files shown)


ReinhardKeil:
We looked at it and think that this will create a conflict with our current way of creating documentation.
The documentation at http://www.keil.com/pack/doc/CMSIS/General/html/index.html is also generated from the headers, but the document structure is supplied in a separate file.

For example for the file cmsis_os.h the documentation (along with the group assignments) are in
https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/DoxyGen/RTOS/src/cmsis_os.txt

Our process for documentation is here: https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/DoxyGen/How2Doc.txt
We use this for all our projects and the concept is to reduce the number of doxygen controls in header files.

bulislaw:
Closing.

PR #161, opened 2017-03-08
Add __TZ_get/set_SP for v8-M

niklas-arm:
Currently missing from v8-M, presumable because there is no msr/mrs for SP for v7-M cos that would be redundant.

You may want to implement this for this for the other compilers too, I'm currently not set up to test that, sorry.

cc @JonatanAntoni

### DIFF
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -320,6 +320,33 @@ __attribute__((always_inline)) __STATIC_INLINE void __TZ_set_MSP_NS(uint32_t top
 #endif
 
 
+#if (defined (__ARM_FEATURE_CMSE ) && (__ARM_FEATURE_CMSE == 3))
+/**
+  \brief   Get Stack Pointer (non-secure)
+  \details Returns the current value of the non-secure Stack Pointer (SP) when in secure state.
+  \return               SP Register value
+ */
+__attribute__((always_inline)) __STATIC_INLINE uint32_t __TZ_get_SP_NS(void)
+{
+  register uint32_t result;
+
+  __ASM volatile ("MRS %0, sp_ns" : "=r" (result) );
+  return(result);
+}
+
+
+/**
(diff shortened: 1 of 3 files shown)


CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -333,6 +333,33 @@ __attribute__((always_inline)) __STATIC_INLINE void __TZ_set_MSP_NS(uint32_t top
...
+  return(result);
+}
+
+
+/**
+  \brief   Set Stack Pointer (non-secure)
+  \details Assigns the given value to the non-secure Stack Pointer (SP) when in secure state.
+  \param [in]    topOfMainStack  Stack Pointer value to set
+ */
+__attribute__((always_inline)) __STATIC_INLINE void __TZ_set_SP_NS(uint32_t topOfStack)
```
JonatanAntoni: Please fix the documentation here. It should be `\param [in]    topOfStack`

JonatanAntoni:
May I ask you to add this functions to `cmsis_armclang.h` as well? Syntax should be the same, so just a matter of copy&paste.

niklas-arm:
Updated.

CMSIS/Core/Include/cmsis_armclang.h
```diff
@@ -320,6 +320,33 @@ __attribute__((always_inline)) __STATIC_INLINE void __TZ_set_MSP_NS(uint32_t top
...
+  register uint32_t result;
+
+  __ASM volatile ("MRS %0, sp_ns" : "=r" (result) );
+  return(result);
+}
+
+
+/**
+  \brief   Set Stack Pointer (non-secure)
+  \details Assigns the given value to the non-secure Stack Pointer (SP) when in secure state.
```
niklas-arm: Though, shouldn't this be `(SP_NS)`? and `(MSP_NS)` etc in the rest of the document? This would align better with the reference manual.

PR #180, opened 2017-04-07
Fix comiplation with Clang for MCU with FPU

vpetrigo:
I have met an issue with using Clang as a front-end for GNU ARM GCC. During compilation I got an issue with `vfpcc` register about which Clang knows nothing:
```make
cmsis\include\cmsis_gcc.h:719:54: error: unknown register name 'vfpcc' in asm __ASM volatile ("VMSR fpscr, %0" : : "r" (fpscr) : "vfpcc", "memory");
```
The fix that I have found is to use `__builtin_arm_set_fpscr()`, but it is unavailable for GCC up to 4.8. So that some preprocessor definitions were used.

Clang version I use:
```bash
$ clang --version
clang version 3.9.1 (tags/RELEASE_391/final)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:\msys64\mingw64\bin
```

GNU ARM GCC version:
```bash
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc.exe (GNU Tools for ARM Embedded Processors 6-2017-q1-update) 6.3.1 20170215 (release) [ARM/embedded-6-branch revision 245512]
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -713,7 +713,11 @@ __attribute__((always_inline)) __STATIC_INLINE void __set_FPSCR(uint32_t fpscr)
 {
 #if ((defined (__FPU_PRESENT) && (__FPU_PRESENT == 1U)) && \
      (defined (__FPU_USED   ) && (__FPU_USED    == 1U))     )
+  #ifdef __clang__
+  __builtin_arm_set_fpscr(fpscr);
+  #else
   __ASM volatile ("VMSR fpscr, %0" : : "r" (fpscr) : "vfpcc", "memory");
+  #endif
 #else
   (void)fpscr;
 #endif


ReinhardKeil:
This change appears incomplete.  What about __get_FPSCR is this working with clang?

vpetrigo:
I have not tested `__get_FPSCR` a lot, but Clang has no issues about it. Clang does not know only about `vfpcc` register.

vpetrigo:
@ReinhardKeil, should I provide you with additional information?

JonatanAntoni:
Hi @vpetrigo. In the meanwhile we got another PR that should fix the issue. May I ask you to verify the latest implementation, please?

vpetrigo:
Hello @JonatanAntoni. Thank you for your reply. I have checked changes made by this 9b5b7561f98cb39f5407959d99c98bdd6eb0fa78 commit and it works perfect. But it seems too strict for GCC (either has `__has_builtin(x)` macro or its version is greater than `7.1`, where that macro for GCC was added). GCC compiler has had appropriate built-ins for working with ARM floating point since [5.4.0](https://gcc.gnu.org/onlinedocs/gcc-5.4.0/gcc/ARM-Floating-Point-Status-and-Control-Intrinsics.html#ARM-Floating-Point-Status-and-Control-Intrinsics). So if you do not mind I can make checks a little bit more permissible.

JonatanAntoni:
Hi @vpetrigo. According to [9b5b756](https://github.com/ARM-software/CMSIS_5/commit/9b5b7561f98cb39f5407959d99c98bdd6eb0fa78) it should work since GCC 5, but also a drawback is given there. Can you please clarify whats about the [patch](https://gcc.gnu.org/ml/gcc-patches/2017-04/msg00443.html)? Does this hurt us? If not we can for sure reduce the restrictions.

nolange:
It does plainly not work with any GCC released and neither with the current WIP, because wrong names were used (`__builtin_arm_ldfscr` and `__builtin_arm_stfscr`).
If the patch for GCC is not applied in time, it will not work with `7.1` but a later version, so the check is already optimistic ;)

vpetrigo:
Hello @JonatanAntoni. Yes, that is my bad, sorry. I have checked out the source and the whole things are in a state, like described in the patch message and @nolange mentioned.

So, to sum up - applied commit https://github.com/ARM-software/CMSIS_5/commit/9b5b7561f98cb39f5407959d99c98bdd6eb0fa78 works perfectly with Clang and current GCC releases. My changes can be dismissed. Thank you for your patience. 😄

JonatanAntoni:
Ok. I close this PR now.

PR #186, opened 2017-04-24
Fix __set_FPSCR and __get_FPSCR for (non-ARM) Clang

nolange:
The clang assembler does not know the vfpcc register, compiling
code using CMSIS will therefor fail.

This patch will use the corresponding builtin thats available on
Clang, and is supposed to be supported with GCC since Version 5.

A patch for GCC 7 is available, fixing a mistake
(https://gcc.gnu.org/ml/gcc-patches/2017-04/msg00443.html).
Optimistically this patch enabled the builtin for GCC 7.1.


nolange:
Hmm, to my defense I searched in the issues but did not check the Pull Requests for similar topics =)

PR #191, opened 2017-04-26
Enable the system header pragma for clang

nolange:
The `#pragma clang system_header` is available since atleast clang 3.3, allow using it

### DIFF
--- a/CMSIS/Core/Include/cmsis_version.h
+++ b/CMSIS/Core/Include/cmsis_version.h
@@ -24,7 +24,7 @@
 
 #if   defined ( __ICCARM__ )
  #pragma system_include         /* treat file as system include file for MISRA check */
-#elif defined (__ARMCC_VERSION) && (__ARMCC_VERSION >= 6010050)
+#elif defined (__clang__) || (defined (__ARMCC_VERSION) && (__ARMCC_VERSION >= 6010050))
   #pragma clang system_header   /* treat file as system include file */
 #endif
 
--- a/CMSIS/Core/Include/core_armv8mbl.h
+++ b/CMSIS/Core/Include/core_armv8mbl.h
@@ -24,7 +24,7 @@
 
 #if   defined ( __ICCARM__ )
  #pragma system_include         /* treat file as system include file for MISRA check */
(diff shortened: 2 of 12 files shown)


JonatanAntoni:
Hi @nolange. Can you please provide me some additional information what you are trying to achieve? I have some concerns that the __clang__ define overrules our intended arm compiler version checks. ARM Compiler 6 is based on Clang and thus __clang__ is (may be) defined here as well.

nolange:
@JonatanAntoni: Quite simply, I want the checks to fit all compilers supporting the pragma.
I can rearrange the checks to `defined (__clang__) && (!defined(__ARMCC_VERSION) || __ARMCC_VERSION >= 6010050)`

Btw, I don`t know what macros ARMCC defines, but I would expect that functionality from the clang base could and should be tested directly instead of ARMs macros?
ie. I would expect to be able to write a check for baseline clang, and the ARM Compiler that`s based on the same clang version would work correctly with it. Seems to be a useless to define the clang macros otherwise.

JonatanAntoni:
Hi @nolange,

Can you point me to the compiler you are using if its not ArmClang, please?
Perhaps we can generalize it completely.

Cheers,
Jonatan

nolange:
Well, its the normal [clang-4.0 from debian](https://packages.debian.org/buster/clang-4.0). It still needs a gcc toolchain for the headers and linking. Invokation is therefore somewhat cumbersome.
`/usr/bin/clang --target=arm-none-eabi --sysroot=/opt/arm-none-eabi-gcc-7.1.0-170509/arm-none-eabi -DSTM32F303xC -DUSE_STDPERIPH_DRIVER  -isystem/opt/arm-none-eabi-gcc-7.1.0-170509/arm-none-eabi/include -mcpu=cortex-m4 -mthumb -mfloat-abi=soft -fshort-enums ....`

JonatanAntoni:
Hi @nolange,

I changed all the system header to simply rely on __clang__. Arm Compiler 6 defines this one as well and thus we should be fine not checking for ARMCC_VERSION in this case.

Can you please double check if my changes fit your expectation?

Thanks,
Jonatan

nolange:
Seems fine, if it works with the (clang-based) ARM Compiler it will work with clang too

PR #210, opened 2017-06-21
Fix return value for osThreadFlagsWait function

bulislaw:
Documentation specifies that osThreadFlagsWait returns flags before
clearing.

http://arm-software.github.io/CMSIS_5/RTOS2/html/group__CMSIS__RTOS__ThreadFlagsMgmt.html#gac11542ad6300b600f872fc96e340ec2b


bulislaw:
After some discussions we agreed that current implementation is correct, but the documentation is a bit misleading (as there can be some side effects changing status while the function executes). Closing.

JonatanAntoni:
I slightly enhanced the documentation with an usage example to explain the return values.
@bulislaw please review commit 67a4147 and close this issue again if no more questions to be discussed.

bulislaw:
Looks good, thanks!

JonatanAntoni:
Finally closing.

PR #212, opened 2017-07-03
Correct gcc version check for __builtin_arm_[gs]et_fpscr builtins

nolange:
My last pull request had some optimisitc version check, the fix was not included in gcc 7.1.

This PR adjusts the version to 7.2, as the gcc-7 branch has the fix included [1], and the next release will be based on that

[1] - https://github.com/gcc-mirror/gcc/commits/gcc-7-branch/gcc/config/arm/arm-builtins.c


JonatanAntoni:
Unfortunately with updating our tests to run with GCC 7.2 reveals that the built-in still has a bug. Thus I need to revert this change.

Issue #129, opened 2017-01-10
_mutex_*  functions are excluded by linker in RTX2 (armcc)


0xc0170:
One of the mbed tests started failing once we updated to RTX2 (mbed malloc tests - https://github.com/ARMmbed/mbed-os/blob/master/TESTS/mbedmicro-rtos-mbed/malloc/main.cpp - using threads, each one malloc some memory). RTX does not fail, so this lead me to believe there is a change in RTX2 that is causing this failure. It's only for ARMCC.

It looks like that ``_mutex_`` wrappers are not in the final image. Looking at the changes between RTX and RTX2, these wrapper functions used to be with `` __attribute__((used))`` to be retained. 
- RTX sources of those wrappers : https://github.com/ARM-software/CMSIS_5/blob/4da2d09bb336e1c08ad54fa9988658eccaa8d169/CMSIS/RTOS/RTX/INC/RTX_CM_lib.h#L338. 
- The RTX2 sources of those wrappers: https://github.com/ARM-software/CMSIS_5/blob/4da2d09bb336e1c08ad54fa9988658eccaa8d169/CMSIS/RTOS2/RTX/Include/rtx_config.h#L531

As I understand these symbols are weakly-linked (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0475m/chr1358938915186.html). And we do not invoke them anywhere, thus I suspect they are weak and set as unused. Debugging shows they are not linked in. if I wrap malloc/free calls plus protect calling their "super" parts with a mutex, the test passes. Or it passes as well if I add "used" attribute.

Any suggestions? Should they be declared with ``__attribute__((used))`` or is there any other fix for this?

0xc0170:
cc @bulislaw @c1728p9

0xc0170:
To reproduce the problem:

Clone mbed-os repository, and run the following command (installed mbed-cli and its test dependencies -  mbed greentea and mbedls and mbedhrun)
```mbed test -n tests-mbedmicro-rtos-mbed-malloc -m K64F -t ARM``.

From master, this test should pass. If you switch to feature_cmsis5 branch on mbed-os repository, the test fails (ERROR reported):

```
+----------+---------------+----------------------------------+----------------------------------+--------+--------+--------+--------------------+
| target   | platform_name | test suite                       | test case                        | passed | failed | result | elapsed_time (sec) |
[...]

PR #213, opened 2017-07-11
Define ARMC mutex operations as weak symbols

bulislaw:
ARMC requires variable number of dynamic mutexes. mbed OS doesn't use dynamic RTX memory pools (carveouts) therefore we need to overwrite the implementation of these.

@JonatanAntoni

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_lib.c
+++ b/CMSIS/RTOS2/RTX/Source/rtx_lib.c
@@ -597,7 +597,7 @@ typedef void *mutex;
 // Initialize mutex
 __USED
 int _mutex_initialize(mutex *m);
-int _mutex_initialize(mutex *m) {
+__WEAK int _mutex_initialize(mutex *m) {
   *m = osMutexNew(NULL);
   if (*m == NULL) {
     osRtxErrorNotify(osRtxErrorClibMutex, m);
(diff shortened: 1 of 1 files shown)


bulislaw:
Ping

RobertRostohar:
I don’t believe this is an acceptable solution. 

Adding __WEAK attribute to those functions which already have the __USED attribute will lead to compilation errors with ARMCC6 (multiply defined symbol) when trying to overwrite the implementation (ARMCC5 does not generate errors). 

Removing the __USED attribute is also not an option (see issue #129).

bulislaw:
That's exactly what @theotherjimmy is seeing.

bulislaw:
Our current solution seems to be:
```
+#if !defined(__ARMCC_VERSION) || __ARMCC_VERSION < 6010050
  __USED
 +#endif
```
could you advise something better?

RobertRostohar:
Removing __USED for ARMCC6 is also not a good solution. It solves only the conflict when overwriting but it won’t tell the linker to keep the weak implementation (see #129).

I don’t have a good generic solution that would satisfy both requirements at this point.

JonatanAntoni:
Hi @bulislaw,

We discussed the issue and came to the conclusion that making our default implementation WEAK does not work. From CMSIS perspective it seems to be a very special requirement for mbedOS no other "normal user" would ever have. Thus we tried to find a less intrusive solution that can be kept under the hood.

What do you think about adding a #ifdef guard? We could introduce a compile time option (e.g. RTX5_NO_LIB_RELOC) which removes the library integration completely if defined.

Cheers,
Jonatan

bulislaw:
It works for us pretty well, thanks!

RobertRostohar:
RTX_NO_MULTITHREAD_CLIB guard has been added which hopefully resolves this issue.

JonatanAntoni:
Hi @bulislaw,

May I ask you to double check if the provided enhancement fits your needs, please?

Thanks,
Jonatan

JonatanAntoni:
We recently integrated this with mbed, thus closing.

PR #226, opened 2017-08-16
Adding IAR files for v8-Baseline

deepikabhavnani:
Assembly files for IAR v8-Baseline support. 

I didn't see support of ifndef in IAR assembly, and CMSIS does not support pre-processing of assembly files. Hence secure and non-secure files are duplicate with just single change "__DOMAIN_NS define".
Let me know if it can be handled in better way.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/IAR/irq_armv8mbl.s
+++ b/CMSIS/RTOS2/RTX/Source/IAR/irq_armv8mbl.s
@@ -1,0 +1,289 @@
+;/*
+; * Copyright (c) 2016-2017 ARM Limited. All rights reserved.
+; *
+; * SPDX-License-Identifier: Apache-2.0
+; *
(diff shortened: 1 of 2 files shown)


JonatanAntoni:
Hi @deepikabhavnani,

pre-processing of assembly files is not a matter of CMSIS. Does IAR support pre-processing like we do for ArmCC and GCC? I think duplicating all the code should only be the very last resort.

deepikabhavnani:
IAR does not assembly equivalent of `.ifndef`, we can use C-style preprocessor` #ifdef` if supported.

JonatanAntoni:
Hi @deepikabhavnani,

May I ask you to review the files IAR recently contributed, please?
Do this files work for you? So we can skip this PR? Or do you want to propose changes?

Thanks,
Jonatan

deepikabhavnani:
Thanks Jonatan, recent IAR files are good to go, will close this PR

PR #229, opened 2017-08-21
Compiler macro for interrupt service routine

AmberAussie:
Hi,

I had trouble with a isr with clang and optimization level 1 on a Cortex-M4 (mk24F12). It seem that it is required to mark it as interrupt. It's strange because http://www.keil.com/appnotes/files/apnt_298.pdf seem to suggest it's optional.

This fix define a general macro for both arm5 and clang compiler to describe a function as ISR.


JonatanAntoni:
Hi @AmberAussie,

thanks for your contribution.

May I ask you to describe the trouble you had in more detail? I would like to understand issue and the way you fixed it. Perhaps you can provide a minimal example that reproduces your issue?

Jonatan

AmberAussie:
Hi,

It occur when I compile my code in O1 with arm compiler v6 (clang). My code is quite simple:

```c
// a function into another module
extern void TimerHandler(void); 

// function from freescale kinetis SDK
void PIT_ClearStatusFlags(PIT_Type *base, pit_chnl_t channel, uint32_t mask)
{
    base->CHANNEL[channel].TFLG = mask;
}

__attribute__((interrupt))
__USED void PIT0_IRQHandler(void)
{
  TimerHandler();
  PIT_ClearStatusFlags(PIT, PIT_CHANNEL_SOFT_TIMER, kPIT_TimerFlag);
}
```

I have no problem if PIT0_IRQHandler has attribute interrupt. But otherwise it affect another part of the code. I think it come to the fact that the compiler optimize function code to not return. If I put directly the line in PIT_ClearStatusFlags into the code of PIT0_IRQHandler I don't have issue. The attribute change the way compiler handle stack and the assembly look different as it save more registers:

CLANG O1 no inerrupt
```
0x00005ED4 B580      PUSH          {r7,lr}
    35:   TimerHandler(); 
    36:    
    37:   /* IT ACK  */ 
0x00005ED6 F002FA45  BL.W          TimerHandler (0x00008364)
    38:   PIT_ClearStatusFlags(PIT, PIT_CHANNEL_SOFT_TIMER, kPIT_TimerFlag); 
    39: } 
0x00005EDA E8BD4080  POP           {r7,lr}
0x00005EDE F000B801  B.W           PIT_ClearStatusFlags (0x00005EE4)
0x00005EE2 0000      MOVS          r0,r0
   243: { 
   244:     base->CHANNEL[channel].TFLG = mask; 
0x00005EE4 F247100C  MOVW          r0,#0x710C
0x00005EE8 2101      MOVS          r1,#0x01
0x00005EEA F2C40003  MOVT          r0,#0x4003
0x00005EEE 6001      STR           r1,[r0,#0x00]
   245: } 
0x00005EF0 4770      BX            lr
```

CLANG O1 inerrupt attribut
```
0x00005ED4 B5D0      PUSH          {r4,r6-r7,lr}
0x00005ED6 AF02      ADD           r7,sp,#0x08
0x00005ED8 466C      MOV           r4,sp
0x00005EDA F36F0402  BFC           r4,#0,#3
0x00005EDE 46A5      MOV           sp,r4
    34:   TimerHandler(); 
    35:  
    36:   /* IT ACK  */ 
0x00005EE0 F002FA46  BL.W          TimerHandler (0x00008370)
    37:   PIT_ClearStatusFlags(PIT, PIT_CHANNEL_SOFT_TIMER, kPIT_TimerFlag); 
0x00005EE4 F000F804  BL.W          PIT_ClearStatusFlags (0x00005EF0)
    38: } 
0x00005EE8 F1A70408  SUB           r4,r7,#0x08
0x00005EEC 46A5      MOV           sp,r4
0x00005EEE BDD0      POP           {r4,r6-r7,pc}
```

Best regards

AmberAussie:
Hi,

You can reject this pull request. I found what was the root cause. This is not compiler issue. It's a timing issue inside the processor.

With optimization we are not lucky. The clear of overflow flag occur and the next instruction is BX LR with LR = 0xFFFFFFFD which make it go out of interrupt. But it seem that the PIT inside the chip hasn't already clear the signal to the core then cpu enter again into interrupt, execute it and exit as this time interrupt was cleared. If I add a NOP between both instruction everything is fine. By adding the attribute((interrupt)) it was also adding more delay so it was fine.

I will check with NXP about this issue but probably this is a silicon/documentation issue.

Best regards

JonatanAntoni:
Hi @AmberAussie,

Too bad, admittedly. This might happen because the actual write to the register (due to buffers and caches) and the propagation through the NVIC logic needs its time. 

You might try to add a Data Synchronization Barrie instruction (__DSB()) to PIT_ClearStatusFlags. This assures that the data is actually written completely before going on.

Please let me know if this works for you. As it is your PR you are welcome to close it after we solved your issues.

Cheers,
Jonatan

JonatanAntoni:
Hi @AmberAussie,

please take note of the [app note](http://www.keil.com/support/docs/3928.htm) discussion your phenomenon as well.

Cheers,
Jonatan

AmberAussie:
Hi,

For info, NXP suggest to do a dummy read of PIT_TCTRLn

Cheers

Issue #167, opened 2017-03-23
RTX primitives object storage sizes


bulislaw:
For non-dynamic allocation, memory can be statically provided using `os*Attr_t` structures. While porting mbed OS to RTX2 I exposed internals of RTX and using `rtx_lib.h` directly declared RTX object eg `os_mutex_t    _obj_mem;` that works fine, but it breaks abstraction layers and reduces portability. The other way would be to count how much memory is needed and declare something like `int32_t _mutex_data[4];` but it doesn't seem to be good solution either.
Do you guys think exposing sizes in `cmsis_os2.h` would be a good idea? We could do it in mbed OS, but I think the same issue would affect other users, if they don't want to have pools of memory statically declared at compile time.
https://github.com/ARMmbed/mbed-os/pull/4011

RobertRostohar:
> I exposed internals of RTX and using  rtx_lib.h 

You should not do that.

`cmsis_o2.h` provides information that is portable.

Implementation specific information (example: control block sizes) should be provided by the implementer.

For example RTX provides the header `rtx_os.h` with all RTX specifics including control block sizes (osRtxMutexCbSize, ...)

kjbracey-arm:
I see that you can't expose sizes in cmsis_os2.h itself - code written using that and dynamic allocation is fully portable, even as a binary library, which is a great achievement.

But in the event that a user is wanting to statically allocate objects, and as such they are accepting that their code will only be compile-time portable between different implementations, shouldn't there be standard naming for a header file defining "current implementation's objects"?

That would gives you two viable models:

    #include "cmsis_os2.h"
    // use dynamic allocation - be totally binary portable

and

    #include "cmsis_os2.h"
    #include "cmsis_os2_storage.h"
    // use static allocation appropriate to current implementation -
    // be compile-time portable

At the minute, rather than the second option, all the code has to do

    #include "cmsis_os2.h"
    #include "rtx_os.h"
    // Use static allocation, and support RTX only. Abstraction not serving its purpose.

RTX's cmsis_os2_storage.h would be basically

    #include "rtx_os.h"
    typedef os_mutex_t osMutexCb_t;
    // etc..

(I prefer using real types like osMutexCb_t to mere size numbers, as that also ensures precise alignment - neither excessive nor inadequate).

[...]

PR #231, opened 2017-08-21
Fix alignment issue with Memorypool implementation - (Fixes issue: ht…

SenRamakri:
Proposed fix for mbed-os issue - https://github.com/ARMmbed/mbed-os/issues/4910. Please review.

Note that there is another pull request for this issue in mbed-os repo - https://github.com/ARMmbed/mbed-os/pull/4941

This fix is little different that it captures the knowledge of alignment in CMSIS in cmsis_os2.h header file. The alignment requirement should not be replicated in different places and should be captured by the platform(in this case its CMSIS) so other clients like MemoryPool.h can use that to get the aligned block size instead of duplicating the alignment knowledge. This also provides better maintenance long term. Adding @c1728p9 as well.

### DIFF
--- a/CMSIS/RTOS/Template/CPP/MemoryPool.h
+++ b/CMSIS/RTOS/Template/CPP/MemoryPool.h
@@ -54,7 +54,7 @@ private:
     osPoolId    _pool_id;
     osPoolDef_t _pool_def;
 #ifdef CMSIS_OS_RTX
-    uint32_t    _pool_m[3+((sizeof(T)+3)/4)*(pool_sz)];
+    uint32_t    _pool_m[OS_MEMORY_POOL_GET_ALIGNED_BLOCK_SIZE(sizeof(T))*(pool_sz)];
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


JonatanAntoni:
Hi @SenRamakri,

Unfortunately your patch is incomplete and mixes up RTX5 internals with CMSIS-RTOS2 inteface. 
Thus we need to discuss about another solution.

First of all I would like to point you to [rtx_os.h](https://github.com/ARM-software/CMSIS_5/blob/401391fc370397eb7eef4cd9d8f49e0f1d80e8b3/CMSIS/RTOS2/RTX/Include/rtx_os.h#L367). There you can find a RTX4 implementation specific macro `osRtxMemoryPoolMemSize` which is exactly what you are looking for, I think.

Additionally we are already discussion how to make all those control block and memory sizes implementation agnostic, please see #167.

Cheers,
Jonatan

SenRamakri:
Closing this PR as the issue has been addressed by a different solution.

Issue #257, opened 2017-10-12
Stack Pointer Issue - Systick handler


deepikabhavnani:
In case of non-secure code, stack pointer was not getting updated properly whenever context switched to another thread.

Issue at https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/RTOS2/RTX/Source/GCC/irq_armv8mbl.S#L232
Instead of LR R7 is used to check if in secure/non-secure mode.

Correct Code
-LSLS     R7,R7,#25              // Check domain of interrupted thread
 -BMI      Sys_ContextSave1       // Branch if secure

+MOV     R0, LR
+LSLS     R0,R0,#25              // Check domain of interrupted thread
+BPL      Sys_ContextSave1       // Branch if non-secure

CC @c1728p9

RobertRostohar:
Thank you for reporting the issue and providing the correction. PR #260 has been merged,

PR #260, opened 2017-10-13
Systick handler switch to secure/nonsecure

deepikabhavnani:
Switch to secure/nonsecure context save/restore is based on 6th bit in LR register, correct the bug (R7 instead of LR was used for decision)

Issue: https://github.com/ARM-software/CMSIS_5/issues/257

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/GCC/irq_armv8mbl.S
+++ b/CMSIS/RTOS2/RTX/Source/GCC/irq_armv8mbl.S
@@ -229,8 +229,9 @@ Sys_ContextSave:
         BL       TZ_StoreContext_S      // Store secure context
         MOV      LR,R7                  // Set EXC_RETURN
         POP      {R1,R2,R3,R7}          // Restore registers
-        LSLS     R7,R7,#25              // Check domain of interrupted thread
-        BMI      Sys_ContextSave1       // Branch if secure
+        MOV      R0,LR                  // Get EXC_RETURN
+        LSLS     R0,R0,#25              // Check domain of interrupted thread
+        BPL      Sys_ContextSave1       // Branch if non-secure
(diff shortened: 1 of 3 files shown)


CMSIS/RTOS2/RTX/Source/GCC/irq_armv8mbl.S
```diff
@@ -229,8 +229,9 @@ Sys_ContextSave:
         BL       TZ_StoreContext_S      // Store secure context
         MOV      LR,R7                  // Set EXC_RETURN
         POP      {R1,R2,R3,R7}          // Restore registers
-        LSLS     R7,R7,#25              // Check domain of interrupted thread
-        BMI      Sys_ContextSave1       // Branch if secure
+        MOV      R0,LR                  // Get EXC_RETURN
+        LSLS     R0,R0,#25              // Check domain of interrupted thread
+        BPL      Sys_ContextSave1       // Branch if secure
```
c1728p9: You might want to update this comment to match the working in context restore. Something like `Branch if non-secure`

c1728p9:
I believe this PR fixes 3 distinct problems. You might want to note these in the commit message:
1. Prevent R7 from being corrupted in Sys_ContextSave
1. Correctly check LR for secure mode
1. Branch when non-secure rather than secure

c1728p9:
@RobertRostohar @JonatanAntoni have you had a chance to look at this? This seems like a fairly big problem for devices with security.

JonatanAntoni:
Hi @c1728p9,

sorry for the delay. We are currently very busy and are going to look into this next week.

Thank you for being patient,
Jonatan

RobertRostohar:
Thank you for detecting and fixing the problem.

Issue #253, opened 2017-10-05
some cosmetic modification in cmsis_armcc.h, cmsis_armclang.h, cmsis_gcc.h, cmsis_iccarm.h


CanastraRF:
Dear Support
I found some difference in the cmsis_armcc.h, cmsis_armclang.h, cmsis_gcc.h, cmsis_iccarm.h

1) different parameter
cmsis_armcc.h, cmsis_armclang.h, cmsis_gcc.h
`int16_t __REVSH(int16_t value)
`
cmsis_iccarm.h
`int32_t __REVSH(int32_t val)
`

2) lower, missing U
 cmsis_iccarm.h
```
 __STATIC_INLINE uint8_t __CLZ(uint32_t data) {
    if (data == 0u) { return 32u; }
                 ^             ^ <-- upper U
    uint32_t count = 0;
    uint32_t mask = 0x80000000;
                              ^  <-- missing U
    while ((data & mask) == 0)
    {
      count += 1u;
                ^                <-- upper U
      mask = mask >> 1u;
    }                 ^          <-- upper U
    return (count);
  }

```

3) different comment to   __REV
cmsis_armcc.h, cmsis_armclang.h
`  \details Reverses the byte order in integer value.
`
cmsis_gcc.h  
`  \details Reverses the byte order in unsigned integer value.
`

4) different comment to __REV16
cmsis_armcc.h, cmsis_armclang.h
`  \details Reverses the byte order in two unsigned short values.
`  
cmsis_gcc.h  
`  \details Reverses the byte order in unsigned short value.
`
Please Change this i neer future
Reto

JonatanAntoni:
Thanks for making us aware of these issues. I'll update the files soon.

TTornblom:
Looking into cmsis_iccarm.h, and comparing it with the definition in https://www.keil.com/pack/doc/CMSIS/Core/html/group__intrinsic__CPU__gr.html#ga1ec006e6d79063363cb0c2a2e0b3adbe, it appears that the definition in cmsis_iccarm.h is correct.

I can fix the "u" vs "U" issues.

TTornblom:
Looks like the CMSIS definition of __REVSH() is wrong, according to this page the types should indeed be int16_t. I'll fix this too.

JonatanAntoni:
Hi @CanastraRF,

I updated all the header files. Can you please double check if
1. According to ACLE it is "int16_t __REVSH(int16_t value)".
[...]

PR #262, opened 2017-10-17
Update for #253

TTornblom:
Looks like the definition of __REVSH differes between http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0662a/CIHCAEJD.html and https://www.keil.com/pack/doc/CMSIS/Core/html/group__intrinsic__CPU__gr.html, and the former is likely the correct one. That is also consistent with the other tool chains.


JonatanAntoni:
I have aligned all the function prototypes in question with ACLE specification. 
May I ask you to double check the latest version, please?
I guess this PR is now obsolete.

TTornblom:
Looks good, closing this PR.

PR #264, opened 2017-10-18
RTOS2/RTX5: Enhanced ThreadTerminate to not allow termination of idle…

JonatanAntoni:
… or timer threads.

Terminating these internal RTX threads leads to undefined behavior and may crash the system.
We should decide whether
- not to expose thread ids of these threads to the outside at all, i.e. through osThreadEnumerate, or
- not to allow termination of these threads.

This patch catches termination of these threads.


JonatanAntoni:
We decided to rather expose thread ids of internal threads. If one really wants to terminate those internal threads it should be possible.

PR #265, opened 2017-10-19
removes unnecessary and non-portable type casts

johan-uden:


llefaucheur:
Hi Johan, thank you for pointing this issue. I will take it into account as part of the next non-regression tasks we have on-going. Thanks again, Laurent.

PR #275, opened 2017-11-04
COSMIC C Compiler does not support recursive inline functions

rajszym:
CSMCC generates errors during compilation:
#error cpcorm ../../Common/CMSIS/include\mpu_armv7.h:191(16) invalid recursive inline function
#error cpcorm ../../Common/CMSIS/include\mpu_armv7.h:195(0) invalid inline function usage
#error cpcorm ../../Common/CMSIS/include\core_cm4.h:2022(0) invalid inline function usage


JonatanAntoni:
Merged in 12314702.

PR #287, opened 2017-12-01
CMSIS-Core(A): Changed __L1C_CleanInvalidateCache function for IAR.

TomoYamanaka:
I changed function by basing on cmsis_gcc.h because the build error(assembly syntax error) will occurs.

### DIFF
--- a/CMSIS/Core_A/Include/cmsis_iccarm.h
+++ b/CMSIS/Core_A/Include/cmsis_iccarm.h
@@ -480,62 +480,77 @@ __IAR_FT void __set_SP_usr(uint32_t topOfProcStack)
 #define __get_mode()                (__get_CPSR() & 0x1FU)
 
 
+__STATIC_FORCEINLINE int32_t log2_up(uint32_t n)
+{
+  int32_t log = -1;
+  uint32_t t = n;
+  while(t)
+  {
+    log++; t >>=1;
+  }
+  /* if n not power of 2 -> round up*/
+  if ( n & (n - 1) ) log++;
+  return log;
+}
+
+__STATIC_INLINE void __L1C_MaintainDCacheSetWay(uint32_t level, uint32_t maint)
+{
+  register volatile uint32_t Dummy;
+  register volatile uint32_t ccsidr;
+  uint32_t num_sets;
+  uint32_t num_ways;
+  uint32_t shift_way;
+  uint32_t log2_linesize;
+  uint32_t log2_num_ways;
+
+  Dummy = level << 1;
+  /* set csselr, select ccsidr register */
+  __set_CCSIDR(Dummy);
+  /* get current ccsidr register */
+  ccsidr = __get_CCSIDR();
+  num_sets = ((ccsidr & 0x0FFFE000) >> 13) + 1;
+  num_ways = ((ccsidr & 0x00001FF8) >> 3) + 1;
+  log2_linesize = (ccsidr & 0x00000007) + 2 + 2;
+  log2_num_ways = log2_up(num_ways);
+  shift_way = 32 - log2_num_ways;
+  for(int way = num_ways-1; way >= 0; way--) {
+    for(int set = num_sets-1; set >= 0; set--) {
+      Dummy = (level << 1) | (set << log2_linesize) | (way << shift_way);
+      switch (maint)
+      {
+        case 0:
+          __ASM volatile("MCR p15, 0, %0, c7, c6, 2" : : "r"(Dummy) : "memory"); // DCISW. Invalidate by Set/Way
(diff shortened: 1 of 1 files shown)


bulislaw:
@JonatanAntoni please review

JonatanAntoni:
Hi @bulislaw,

I already commented this part of the code: It seems not to be specific to IAR so why is it placed in cmsis_iccarm.h? Can we refactor the cache handling to become compiler agnostic?

Cheers,
Jonatan

bulislaw:
@TomoYamanaka ping

TomoYamanaka:
I developed the codes that refactor the cache handling as compiler agnostic in my temporary branch.
https://github.com/TomoYamanaka/CMSIS_5/tree/feature_refactor_cache_handling
I am now focusing on Mbed, compile and test for Cortex-A were successfully by using ARMCC(ver5), GCC(ver6.2) and IAR(ver8.20.1).

Could you review this changes? If there is no problem with the contents, I will give this contents to you as another PR, and close this PR.
Or please let me know if you prefer another review way.

JonatanAntoni:
Hi @TomoYamanaka,

I think the changes are pretty much what we should go for. 

We should not remove `__L1C_CleanInvalidateCache` completeley but keep a stub for compatibility reasons. The stub can be flagged as deprecated and should not be used, i.e. with `__attribute__((deprecated))`.

Another issue I came across while testing the routine is the usage of `L1C_CleanInvalidateCache(0)`. The call stack might get corrupted if this function is not really inlined, e.g. due to low optimization. The values pushed to the stack are by chance not written back before we invalidate the cache content. Thus garbage is restored on function return.

In order to get at least an educated guess whether the implementation does the job we should enhance [CoreValidation](https://github.com/ARM-software/CMSIS_5/tree/develop/CMSIS/CoreValidation) tests. Do you have an idea how we could implement some sort of self-test routine for cache maintenance?

Thanks,
Jonatan

JonatanAntoni:
Hi @TomoYamanaka,

I added this stuff with commit 17b53ef. May I ask you to give it a review, please?

I created some [CoreValidation tests](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/CoreValidation/Source/CV_L1Cache.c) as well. Unfortunately I face some test failures on various configurations, i.e. using IAR and on Cortex-A7. We need to fix these, soon.

Cheers,
Jonatan

TomoYamanaka:
Hi @JonatanAntoni 

> I added this stuff with commit 17b53ef. May I ask you to give it a review, please?

I will review the changes of part of `CMSIS/Core_A/Include/` in commit 17b53ef.
Please give me a little more time.

But sorry, since I can not review the part of `CMSIS/CoreValidation/`, could you please assign other members?

TomoYamanaka:
I reviewed the changes of part of CMSIS/Core_A/Include/ in commit 17b53ef, but I guess that there is a problem in only GCC.
I am now focusing on Mbed, the build error will occur in GCC toolchain only. 
The spot of build error is [here](https://github.com/ARM-software/CMSIS_5/commit/17b53ef2a297ef73d0fb0d720cf2a3458d6127e2#diff-6ebf31aa0077600595814f0e3a7bbaa6R1013), error message is `expected unqualified-id before numeric constant`.
It seems that GCC compiler work as a reserved word of  `__DEPRECATED = 1`.
Reference is [here](\GNU Tools ARM Embedded\6 2017-q2-update\share\doc\gcc-arm-none-eabi\html\cpp\Common-Predefined-Macros.html).
```
__DEPRECATED
This macro is defined, with value 1, when compiling a C++ source file with warnings about deprecated constructs enabled. These warnings are enabled by default, but can be disabled with -Wno-deprecated. 
```
<br>
As a test, when I inserted the below code into the 58th line of cmsis_gcc.h, the build was successful.

`#undef   __DEPRECATED`

JonatanAntoni:
Thx @TomoYamanaka,

We need to rename `__DEPRECATED` to something else in order to prevent that name clash.
What do you think about replacing it with `CMSIS_DEPRECATED`?

Cheers,
Jonatan

TomoYamanaka:
Hi @JonatanAntoni 
Thank you for your comments. 
Your suggestion looks good to me.

TomoYamanaka:
@JonatanAntoni 
Thank you for the update.
Since my PR was replaced to the below commits as a permanent fix, I close this PR.
- https://github.com/ARM-software/CMSIS_5/commit/17b53ef2a297ef73d0fb0d720cf2a3458d6127e2
- https://github.com/ARM-software/CMSIS_5/commit/d43938fac6e0fa1b5682c2ca5704277b6dc34963

PR #288, opened 2017-12-01
RTX5: Added __get_PSP function for IAR

TomoYamanaka:
The build error(syntax error) will occurs because assembler instructions of IAR is different from GCC.
Therefore, I added function for IAR by using "#elif defined(__ICCARM__)".

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_core_ca.h
+++ b/CMSIS/RTOS2/RTX/Source/rtx_core_ca.h
@@ -112,7 +112,23 @@ static __asm    uint32_t __get_PSP (void) {
   sub   r0, r0, #32
   bx    lr
 }
-#else
+#elif defined(__ICCARM__)
+__STATIC_INLINE __arm uint32_t __get_PSP (void) {
+  register uint32_t ret;
+
+  __asm volatile (
+    "sub  sp,sp,#4\n\t"
+    "stm  sp,{sp}^\n\t"
+    "pop  {%[ret]}\n\t"
+    "sub  %[ret],%[ret],#32\n\t"
+    : [ret] "=&l" (ret)
+    :
+    : "memory"
+  );
+
+  return ret;
+}
+#else   // !(defined(__CC_ARM) || defined(__ICCARM__))
 __STATIC_INLINE uint32_t __get_PSP (void) {
   register uint32_t ret;
 


JonatanAntoni:
Can we reduce the portion of duplicated code in this part to the absolute minimum, please?

RobertRostohar:
What is different for IAR vs GCC. Existing GCC could should be ok.

JonatanAntoni:
I think its only because 
>      ".syntax unified\n\t"
>      ".arm\n\t"
is not valid for IAR. There are already ifdefs around. Looks like mbed was using an older version.

@bulislaw can you please double check if the latest version already works for you?

TomoYamanaka:
In IAR of mbed environment, 

> I think its only because
> 
> ```
> ".syntax unified\n\t"
> ".arm\n\t"
> ```
> 
> is not valid for IAR.

In addition above, it seems that build will be successful if add **__arm** in function type.([reference](https://github.com/ARM-software/CMSIS_5/pull/288/commits/4bff5048709f603631e7b8d97efa6df862dbeb39#diff-fbc2a7245b9471b77dddd09091ff0559R116))

JonatanAntoni:
We are running tests using IAR 8.20.1 and they compile successfully.
Do we really need to append __arm to the function signature for IAR?
Can you please check the latest version we have in RTX. This already contains a preprocessor condition for ICCARM.

TomoYamanaka:
> Do we really need to append __arm to the function signature for IAR?

At least, in mbed environment that I have based on, build error occured when didn't add append **__arm**.
```
[Error] rtx_core_ca.h@85,15: [Og006]: Error in inline assembly: "Operand syntax error"
[Error] rtx_core_ca.h@85,0: [Og006]: Error in inline assembly: "Operand syntax error"
[ERROR]
      "stm  sp,{sp}^\n\t"
               ^
```


I would like to let me check one point,
> IAR 8.20.1

is it EWARM ver8.20.1?


> Can you please check the latest version we have in RTX. This already contains a preprocessor condition for ICCARM.

Is it `CMSIS_5\CMSIS\RTOS2\RTX\Source\` ?
Sorry, there are too many change amounts against mbed environment so I can not handle it.

TomoYamanaka:
> We are running tests using IAR 8.20.1 and they compile successfully.

Does  this test success contain Cortex-A?
If it is right, I would like to confirm it too. Could you please let me know how to running test by using CMSIS repo, Cortex-A and IAR?

JonatanAntoni:
@TomoYamanaka the tests are not publicly available.
How do you compile for Cortex-A, typically? For me it looks like you're trying to use Thumb instruction set by default? And thus you might beto force arm instruction set for this function?

TomoYamanaka:
Yes.
In IAR of my environment, Cortex-A is Thumb instruction set by default.
But it seems that this __get_PSP function is made Arm instruction set, so I added "__arm" and forced to set Arm instruction.

JonatanAntoni:
Okay. Now I got the point, I guess.
Do you need (or want) to build Cortex-A code with Thumb instructions?
You might switch to use Arm instructions which is the default afaik. 
If you have specific requirements for Thumb on Cortex-A we should take a closer look into because we didn't consider this at all, yet.

TomoYamanaka:
Yes. We are now focusing on Mbed, and Mbed build Cortex-A code with Thumb instruction by default.
From this reason, this modification is needed shortly in order to make our platform be Mbed enabled.

bulislaw:
@TomoYamanaka considering that Renesas board is the only Cortex A board in Mbed it's not fair to say we build it using Thumb. I'm pretty sure everything is open for discussion.

RobertRostohar:
`__get_PSP` function has been updated (slightly nicer solution - less duplication) which should fix the problem.

TomoYamanaka:
Since `__get_PSP` function was updated by [this commit](https://github.com/ARM-software/CMSIS_5/commit/af5c8c4652ca6803ef78ed47c4736a44ad27f12d), I close this PR.

PR #304, opened 2018-01-13
Core(M): Fix -Wundef for __ARM_FEATURE_DSP

salkinium:
This prevents multiple warnings when compiling with GCC
```
ext/cmsis/cmsis_gcc.h:1539:6: warning: "__ARM_FEATURE_DSP" is not defined, evaluates to 0 [-Wundef]
 #if (__ARM_FEATURE_DSP == 1)                             /* ToDo ARMCLANG: This should be ARCH >= ARMv7-M + SIMD */
      ^~~~~~~~~~~~~~~~~
```

I also fixed the same issue for ICC.

@JonatanAntoni: I believe the `ToDo` does not apply anymore, since you now have a separate header for armclang anyways. I took the liberty of removing it, is that correct?

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -1536,7 +1536,7 @@ __STATIC_FORCEINLINE uint32_t __STLEX(uint32_t value, volatile uint32_t *ptr)
   @{
 */
 
-#if (__ARM_FEATURE_DSP == 1)                             /* ToDo ARMCLANG: This should be ARCH >= ARMv7-M + SIMD */
+#if (defined (__ARM_FEATURE_DSP) && (__ARM_FEATURE_DSP == 1))  /* ToDo ARMCLANG: This should be ARCH >= ARMv7-M + SIMD */
 
 __STATIC_FORCEINLINE uint32_t __SADD8(uint32_t op1, uint32_t op2)
 {


JonatanAntoni:
Thanks @salkinium,

the additional checks looks viable to me. 

Can we cross check with ARM Compilers, please?
- In armclang we check for `__ARM_ARCH_7EM__` to detect SIMD instructions.
- In armcc we map `__TARGET_ARCH_7E_M` to `__ARM_ARCH_7EM__`.
- In ACLE 2.1 `__ARM_FEATURE_DSP` is documented.

Best,
Jonatan

salkinium:
> Can we cross check with ARM Compilers, please?

I'm afraid I don't fully understand what you mean.
You mean adding aditional checks like these in `core_cm33.h`?
```c
  #if defined (__ARM_FEATURE_DSP) && (__ARM_FEATURE_DSP == 1U)
    #if defined (__DSP_PRESENT) && (__DSP_PRESENT == 1U)
      #define __DSP_USED       1U
    #else
      #error "Compiler generates DSP (SIMD) instructions for a devices without DSP extensions (check __DSP_PRESENT)"
      #define __DSP_USED         0U    
    #endif
  #else
    #define __DSP_USED         0U
  #endif
```

Even so, I just want the warning to go away, since it spams my compile output. I've refactored to keep the ToDo comment, and removed the ICC fix, so that this only fixes this warning and doesn't change anything else.

JonatanAntoni:
I checked `__ARM_FEATURE_DSP` for GCC and IAR and it should be fine.
Thanks for contributing.

PR #305, opened 2018-01-15
Update cmsis_os.h

JohannesTrageser:
If compiled with ARM Compiler V6 a warning is generated (see Log Output below)
Adding an #ifndef __NO_RETURN to cmsis_os.h resolves this problem.

C:/Keil/Keil_v5/ARM/PACK/ARM/CMSIS/5.2.0/CMSIS/RTOS/RTX/INC\cmsis_os.h(73): warning: '__NO_RETURN' macro redefined [-Wmacro-redefined] #define NO_RETURN attribute__((noreturn))
C:/Keil/Keil_v5/ARM/PACK/ARM/CMSIS/5.2.0/CMSIS/Include/cmsis_armclang.h(50): note: previous definition is here
  #define NO_RETURN                            attribute__((__noreturn__))
1 warning generated.

### DIFF
--- a/CMSIS/RTOS/RTX/INC/cmsis_os.h
+++ b/CMSIS/RTOS/RTX/INC/cmsis_os.h
@@ -67,6 +67,7 @@
 #define os_InRegs
 #endif
 
+#ifndef __NO_RETURN
 #if   defined(__CC_ARM)
 #define __NO_RETURN __declspec(noreturn)
 #elif defined(__ARMCC_VERSION) && (__ARMCC_VERSION >= 6010050)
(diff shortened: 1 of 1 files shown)


JonatanAntoni:
Thanks for pointing that out.

Please not that this cmsis_os.h belongs to RTX4 which has been superseded with RTX5 a while ago. Perhaps you want to switch to CMSIS-RTOS2 API using RTX5 which is still under active maintenance.

Best,
Jonatan

JonatanAntoni:
Hi @JohannesTrageser,

Do you have any objections to switch to RTX5?
We'd rather like to keep RTX4 as it is only for historical reasons. We expect all up-to-date projects to rely on RTX5 which is still under active development and maintenance. It has a couple of benefits and should be preferred nonetheless.

Best,
Jonatan

PR #316, opened 2018-02-09
Rtx5: inhibit 'unused parameter' warning when RtxTimerThread is empty…

pbl-pw:
… and compile with '-Wextra' flag

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_lib.c
+++ b/CMSIS/RTOS2/RTX/Source/rtx_lib.c
@@ -219,7 +219,7 @@ static const osMessageQueueAttr_t os_timer_mq_attr = {
 #else
 
 extern void osRtxTimerThread (void *argument);
-       void osRtxTimerThread (void *argument) {}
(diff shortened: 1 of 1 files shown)


JonatanAntoni:
Hi @pbl-pw,

thanks for your contribution.
Unfortunately this way of defining the parameter as "intentionally unused" is not portable. `__attribute__((unused))` is a GNUC extensions and not C standard.

In other places we add: `(void)argument;` to the function body to silence the compiler.

Another option would be to enhance CMSIS-Core `compiler.h` with a new __UNUSED(x) macro, like discussed on [Stack Overflow](https://stackoverflow.com/questions/7090998/portable-unused-parameter-macro-used-on-function-signature-for-c-and-c). We need to assure that such a macro would be valid for a wide range of compilers.

Best,
Jonatan

RobertRostohar:
Should be fixed now in a compiler agnostic way.

JonatanAntoni:
@pbl-pw, may I ask you to double check the fix, please?
Please close this PR if the fix solves your issue. Thanks.

PR #317, opened 2018-02-09
Added workspace and project files for DSP on IAR.

TTornblom:


CMSIS/Core/Include/core_cm23.h
```diff
@@ -92,7 +92,7 @@
 
 #elif defined ( __ICCARM__ )
   #if defined __ARMVFP__
-    #error "Compiler generates FPU instructions for a device without an FPU (check __FPU_PRESENT)"
+    #warning "Compiler generates FPU instructions for a device without an FPU (check __FPU_PRESENT)"
```
JonatanAntoni: This is now unaligned with the other compilers. Can you please give a short justification why usage of FPU instructions on a device without FPU should be a warning instead of an error?
I expect images created with wrong FPU settings to instantaneously crash at run-time, i.e. as soon as the first FPU instruction is hit.

TTornblom:
I had a chat with the engineer that did these changes, and it turns out that we have patched the files for m4 and m7 like this since 2014. His recollection was that there are (were?) some examples that didn't compile otherwise, cause unknown.

The engineer also said that in version 4 it used to be #warning, but was changed to #error, and checking the history it seems to have happened between 4.30 and 4.40.

Apparently customers started complaining about getting compile errors due to this and we started patching it back.

Looking at the files I notice that ARMCC has the same #warning for m33, m4 and m7. These are the only targets that seems to have an FPU option, so we will revert m3 and m23 as they were, and only change m4, m7 and m33.

JonatanAntoni:
Hi @TTornblom,

I would highly appreciate if we can figure out the actual issue you try to circumvent.
For sure CM0, CM0+, CM3 and CM23 do not have a FPU option. Thus compiling code for these core with FPU instructions seems always to be a mistake. 

On CM4, CM7 and CM33 the FPU is optional and it depends on the actual device used. In CMSIS each device should have a device header file, like [ARMCM4_FP.h](https://github.com/ARM-software/CMSIS_5/blob/develop/Device/ARM/ARMCM4/Include/ARMCM4_FP.h) for a generic Cortex-M4 device. The define `__FPU_PRESENT` tells us whether FPU is available or not. Code compiled with FPU instructions will crash at runtime on devices without an FPU. The preprocessor error should prevent weird runtime effects.

In order to have CMSIS aligned for all toolchains we should agree on a common behavior. What shall happen if one compiles code using FPU instructions for a target device without FPU?

Thanks,
Jonatan

JonatanAntoni:
Hi @TTornblom,

I reviewed your changes and tried to apply them to the latest develop.
If you don't mind could you please clean-up your commits, slightly?
 - Remove the commits completely that are simply add and revert.
 - Prefix the two relevant commits (03d5486 and 0f3db9c) with the affected component, i.e. "DSP:" or "NN:", respectively.
 - The examples added to NN are currently located under an ARM folder. Wouldn't it be easier to understand to separate the examples into an IAR folder, completely? I guess we should reference the examples from the PDSC file as well.

Cheers,
Jonatan

TTornblom:
No longer relevant

Issue #345, opened 2018-04-02
support removal of register storage class specifier in C++17


pabigot:
Prior to C++17 the `register` storage class specifier was deprecated; in C++17 it has been removed, making attempts to compile with CMSIS 5.3 with gcc `--std=c++17` fail.  Something like the following is needed in every header that uses that feature, along with the corresponding modifications at all use-points.

    #if (__cplusplus - 0) >= 201703L
      #define __REGISTER
    #else
      #define __REGISTER                             register
    #endif

See: http://en.cppreference.com/w/cpp/language/storage_duration

JonatanAntoni:
May I ask you to be slightly more specific, please?
Where exactly do you have this issue?

Please be aware that CMSIS itself it intended to be ANSI-C (C99) and not C++. Simply removing the storage class `register` from the code is not guaranteed to work. Depending on the implementation it might be required that a value is kept as a register value only.

Using CMSIS from C++ should be possible when compiling all .c files in C-Mode. Inline functions defined in header files should be compiled as `extern C`. Please let me know if that does not work for you.

pabigot:
As of C++17 the existing solution does not work.

CMSIS Core headers like `core_cm0.h` can be included into C++ source files; this is enabled through the use of the `__cplusplus` macro to add the required `extern "C"` block around the declarations.  This was sufficient to allow core inline functions like `NVIC_SetEnableIRQ()` and structures like `SCB_Type` to be available to and used in C++ applications through C++14.

The change to the `register` storage class specifier in C++17 means this isn't possible anymore because simply including `core_cm0.h` into a C++ application pulls in `cmsis_compiler.h` which pulls in `cmsis_gcc.h` which declares a local `register uint32_t result;` which is a syntax error.  The modification I suggest is required if the existing capability is to be maintained beyond C++14.

The change can probably be conditional on using C++ at an affected version (as suggested) and isolated to compiler-specific headers like `cmsis_gcc.h` which should instead declare `__REGISTER uint32 result;`. For implementation files that use register declarations in C99-compiled code I agree it's not relevant and undesirable.

Some information on the reason for the change is [here](https://stackoverflow.com/a/30809775/236848). As that material indicates gcc completely ignores the `register` keyword unless all optimizations are explicitly disabled (the `__ASM` statements in CMSIS include a constraint that ensures a register is used).  In practice I've had no problems with gcc-arm-none-eabi-7-2017-q4-major and `--std=c++17` with this modification.

[...]

PR #347, opened 2018-04-10
fixes to support C++17 in mixed-language headers

pabigot:
First commit satisfies the basic needs of issue #345.  These modifications have been tested with gcc-arm-none-eabi-7-2017-q4-major (GCC 7.2.1) with c++14 and c++17.  They've also been made for armclang, but I have not tested that toolchain.

The second commit generalizes it slightly, to cover the use of the deprecated/removed qualifier in the cm7 header (the only other appearance of the qualifier in the Core headers).  I have no cm7 code so have not been able to test this change.

### DIFF
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -43,9 +43,9 @@
 #ifndef   __STATIC_INLINE
   #define __STATIC_INLINE                        static __inline
 #endif
-#ifndef   __STATIC_FORCEINLINE                 
+#ifndef   __STATIC_FORCEINLINE
   #define __STATIC_FORCEINLINE                   __attribute__((always_inline)) static __inline
-#endif                                           
+#endif
 #ifndef   __NO_RETURN
   #define __NO_RETURN                            __attribute__((__noreturn__))
 #endif
@@ -109,6 +109,11 @@
 #endif
 #ifndef   __RESTRICT
   #define __RESTRICT                             __restrict
+#endif
+#if (__cplusplus - 0) >= 201703L
+  #define __REGISTER
+#else
+  #define __REGISTER                             register
 #endif
 
 
@@ -237,7 +242,7 @@ __STATIC_FORCEINLINE uint32_t __get_xPSR(void)
  */
 __STATIC_FORCEINLINE uint32_t __get_PSP(void)
 {
-  register uint32_t result;
+  __REGISTER uint32_t result;
 
   __ASM volatile ("MRS %0, psp"  : "=r" (result) );
   return(result);
@@ -252,7 +257,7 @@ __STATIC_FORCEINLINE uint32_t __get_PSP(void)
  */
 __STATIC_FORCEINLINE uint32_t __TZ_get_PSP_NS(void)
 {
-  register uint32_t result;
+  __REGISTER uint32_t result;
 
   __ASM volatile ("MRS %0, psp_ns"  : "=r" (result) );
   return(result);
@@ -291,7 +296,7 @@ __STATIC_FORCEINLINE void __TZ_set_PSP_NS(uint32_t topOfProcStack)
  */
 __STATIC_FORCEINLINE uint32_t __get_MSP(void)
 {
-  register uint32_t result;
+  __REGISTER uint32_t result;
 
   __ASM volatile ("MRS %0, msp" : "=r" (result) );
   return(result);
@@ -306,7 +311,7 @@ __STATIC_FORCEINLINE uint32_t __get_MSP(void)
  */
 __STATIC_FORCEINLINE uint32_t __TZ_get_MSP_NS(void)
 {
-  register uint32_t result;
+  __REGISTER uint32_t result;
 
   __ASM volatile ("MRS %0, msp_ns" : "=r" (result) );
   return(result);
@@ -346,7 +351,7 @@ __STATIC_FORCEINLINE void __TZ_set_MSP_NS(uint32_t topOfMainStack)
  */
 __STATIC_FORCEINLINE uint32_t __TZ_get_SP_NS(void)
 {
-  register uint32_t result;
+  __REGISTER uint32_t result;
 
   __ASM volatile ("MRS %0, sp_ns" : "=r" (result) );
   return(result);
@@ -570,7 +575,7 @@ __STATIC_FORCEINLINE void __TZ_set_FAULTMASK_NS(uint32_t faultMask)
   Devices without ARMv8-M Main Extensions (i.e. Cortex-M23) lack the non-secure
   Stack Pointer Limit register hence zero is returned always in non-secure
   mode.
-  
+
   \details Returns the current value of the Process Stack Pointer Limit (PSPLIM).
   \return               PSPLIM Register value
  */
@@ -581,7 +586,7 @@ __STATIC_FORCEINLINE uint32_t __get_PSPLIM(void)
     // without main extensions, the non-secure PSPLIM is RAZ/WI
   return 0U;
 #else
-  register uint32_t result;
+  __REGISTER uint32_t result;
   __ASM volatile ("MRS %0, psplim"  : "=r" (result) );
   return result;
 #endif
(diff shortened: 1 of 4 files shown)


CMSIS/Core/Include/cmsis_compiler.h
```diff
@@ -261,6 +261,12 @@
   #error Unknown compiler.
 #endif
 
+/* Support the register storage class specifier in mixed-language
+ * headers that might be referenced when using a C++17 or later
+ * compiler. */
+#ifndef __REGISTER
+#define __REGISTER register
```
JonatanAntoni: Do you think this is needed at all? As stated in our discussion only parts of the functions are currently using the storage class specifier at all. Perhaps we can drop it, completely?

pabigot: I think it is needed.

CMSIS uses in C and C++ with gcc have had the specifier in the past.  I don't know what its presence or absence might imply to different compiler versions and optimization settings.  The stackoverflow comment included:

> The register keyword does still have a use, though, in a gcc
> extension: gcc uses it in combination with asm to implement register
> variables.

This may refer just to things like:

    register uint32_t __regProcessStackPointer  __ASM("psp")

or it may mean that in:

    __REGISTER uint32_t result;
    __ASM volatile ("MRS %0, psp"  : "=r" (result) );
    return(result);

the presence of the qualifier ensures that stack copying doesn't occur, outside the code generated by the __ASM statement because the constraint requires it.  My understanding of `-O0` is that if `register` isn't present gcc compiling plain C probably will copy into and out of a stack-allocated definition of `result`, in which case removing the qualifier will change the generated code.

So I see no value, and potential risk, in removing the qualifier in any situation where it has always been there and is allowed.  It's only in C++17 that it **cannot** appear so must be removed.

thopre01: Hi,

register keyword is just a hint to the compiler [1], it is free to ignore it. I do not know for sure but it is very possible that GCC ignores it at -O0. Besides, while it may indeed change the code generation, removing the keyword makes the CMSIS source code simpler so I think it is the right approach here.

[1] C99 6.7.1 says "A declaration of an identifier for an object with storage-class specifier register
suggests that access to the object be as fast as possible. The extent to which such
suggestions are effective is implementation-defined."

Regarding the stack overflow comment, it refers indeed to constructs such as `register uint32_t __regProcessStackPointer  __ASM("psp")` which is not what CMSIS uses.

So in summary: while removing register *might* change the code generated it will not break the code and will make the source simpler.

Best regards.

pabigot: @thopre01 If you want to submit an alternative PR that removes it entirely, feel free.  I'm not going to take the chance of breaking working code when it's trivial to avoid doing so.

JonatanAntoni: @pabigot Thomas (thopre01) is a GCC expert in the team, here. I just asked him to clarify the need of `register` specifier. As it seems not to be valuable nor needed I'd rather remove this bit completely instead of adding more compatibility stuff. Furthermore `register` keyword is not used consistently at all.

pabigot: Fair enough.  My position is that having used gcc for over 25 years (and dealt with the effects of upgrades/compiler changes), and having maintained an out-of-tree back end for a few years as well, I won't sign off on a change that removes the qualifiers entirely, especially for embedded applications where how the compiler generates instructions for a given code sequence is sometimes critical.  Thomas is on the team, so he's more qualified to make that change and deal with any fallout.  If you prefer that approach just close this PR unmerged.

pabigot: FWIW: my test with arm-gcc shows `-O0` does generate different instructions in C code when `register` is removed.

thopre01: But then you would have a change whenever the compiler switches to C++17 or later by default (which is arguably in a long time) anyway. Besides, I would expect a user who have code size or speed requirement to use -Os/-O2/-O3/-Ofast.

Anyway there is a tradeoff here. Adding a compatibility option creates more codepath which means you need more testing for any change done afterwards or you risk breaking one of the codepath. In the long term I would think removing the qualifier would be better.

JonatanAntoni: I tried to reproduce this myself. Using GCC 7.2.1 with -O0 in C99 mode seems to recognize `register` specifier and produces more efficient code than without the keyword. Starting at -O1 the compiler optimizes both version into the same result, i.e. not using a intermediate stack variable at all.

So the decision is slightly different than stated before. Do we want to have the same optimal code in C mode for all optimization levels? Do we accept the slightly more complex code to be maintained? If yes shall we update all the intrinsic functions to use the specifier to be consistent?

I tend to argue that having optimal code should not be a matter of interest when using -O0. But to be fair this might depend on the actual situation. Removing the specifier would cleanup the code which looks like a benefit on the long run to me. As the downside we affect the code generation for existing projects slightly, i.e. having two additional instructions and more stack usage.

Concluding this I would still remove the specifier completely. We are used to stay API compatible between CMSIS releases. But constraining us to be binary equivalent looks like a path into a dead end.

@pabigot does it work for your applications if we drop usage of `register` specifier?

JonatanAntoni: @0xc0170: Can you review this C++ issue from mbed perspective?

pabigot: @JonatanAntoni For the code I'm currently working on the register specifiers are removed by my patch, so there should be no difference if a different patch that unconditionally removes them is merged instead.  I'm not currently supporting any non-C++ code where I would be affected if the removal has a visible effect.

Keep in mind that core_cm7.h has register qualifiers in code shared by all compilers, so removing it in GCC-only headers does not fully resolve issue #345.  As I recall that's the only other instance in mixed-language headers, so possibly they could be removed there as well, but then you'd have to see whether IAR and other vendors also ignore the qualifier.

CMSIS/Core/Include/cmsis_armclang.h
```diff
@@ -43,9 +43,9 @@
 #ifndef   __STATIC_INLINE
   #define __STATIC_INLINE                        static __inline
 #endif
-#ifndef   __STATIC_FORCEINLINE                 
+#ifndef   __STATIC_FORCEINLINE
   #define __STATIC_FORCEINLINE                   __attribute__((always_inline)) static __inline
-#endif
```
0xc0170: formatting should be separate commit, distracts from real changes

CMSIS/Core/Include/cmsis_compiler.h
0xc0170: Looks fine to me to be removed completely in the codebase

JonatanAntoni: @pabigot Thanks for pointing that detail out, again. Let me think about all the consequences again.

FreddieChopin:
If that matters at all, I would be for complete removal of "register" keyword too. It was deprecated in C++ because compilers rarely use it for anything useful - asm constraints are more than enough to have the variable in the register (or at least copied to/from register). Trying to optimize a `-O0` case is not a good reason for leaving this keyword - the marginal size increase it may bring in that case will most likely be several orders of magnitude below increase caused by other parts of project, if any at all. I've even tested that with a project I have on my desk - at `-O0` the size with/without `register` is identical.

I may try to provide an alternate PR too.

JonatanAntoni:
Thanks @FreddieChopin,

I discussed this issue with a couple of compiler experts. The conclusion is that removing `register` completely for GCC and ArmClang should not affect the functionality. In some cases the code efficiency at `-O0` might be slightly worse. This should be acceptable as `-O0` means "non optimized". 

We must not remove the `register` keyword used for Arm Compiler 5. In ARMCC this keyword combined with `asm` keyword is used to declare named register variables, i.e. to access special function registers.

IAR Compiler is not affected at all, i.e. no `register` keyword used in `cmsis_iccarm.h`.

It is very appreciated if you want to provide the needed changes as a pull-request.

FreddieChopin:
@JonatanAntoni pull request which removes the keyword wherever possible - #365

pabigot:
@JonatanAntoni @FreddieChopin By inspection it should work fine since it removes the declaration completely.  However I won't be back on ARM development for another week or two at earliest; I will try to check it then.  In any case this PR should be closed.

PR #360, opened 2018-05-08
Split rfft_fast_init_f32 into separate initializer functions.

dflogeras:
The rationale here is that if the caller knows the size FFT they require
at compile time, only the look-up tables required for that size are
placed in the data segment (ROM).  This saves between ~40k to ~80k of
ROM when only using a single FFT size.

### DIFF
--- a/CMSIS/DSP/Include/arm_math.h
+++ b/CMSIS/DSP/Include/arm_math.h
@@ -2224,6 +2224,23 @@ arm_status arm_rfft_fast_init_f32 (
    arm_rfft_fast_instance_f32 * S,
    uint16_t fftLen);
 
+arm_status arm_rfft_32_fast_init_f32 ( arm_rfft_fast_instance_f32 * S );
+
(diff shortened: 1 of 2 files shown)


JonatanAntoni:
That sounds reasonable to me at the first glance. We need to review this carefully. I guess it might be worth to refactor old `arm_rfft_fast_init_f32` function to use the new ones to reduce duplicate code.

dflogeras:
I have written a separate patch to re-implement the old function in terms of the newly added ones.  I'll submit another PR when this one is done review.

llefaucheur:
arm_rfft_4096_fast_init_f32 (line223) should be 4096 instead of 1024.

This patch allows to create smaller programs. The current switch/case do not help the compiler to decide which ROM table to keep. Thanks

dflogeras:
Ah, good catch thank you.  How shall we proceed?  Do I redo my patches and re-submit the PR?

llefaucheur:
If you don't mind, I would prefer you re-submit. Thanks.

dflogeras:
Resubmitted as #369

PR #369, opened 2018-05-15
Split rfft ctors

dflogeras:
Supersedes PR #360


llefaucheur:
Hi, Thanks for coming back to us, this issue will be considered.
We are currently considering other options, and, for short-term, prefer focusing on the big picture of the issue you pointed to. Regards, Laurent.

PR #376, opened 2018-06-07
make write to CP volatile so it doesn't get optimised out

ithinuel:
If reading from these registers can have side effect, I will add the volatile keyword to the `__get_CP` macro too.


CMSIS/Core_A/Include/cmsis_armcc.h
```diff
@@ -452,7 +452,7 @@ __STATIC_INLINE void __set_FPEXC(uint32_t fpexc)
  */
 
 #define __get_CP(cp, op1, Rt, CRn, CRm, op2) do { register uint32_t tmp __ASM("cp" # cp ":" # op1 ":c" # CRn ":c" # CRm ":" # op2); (Rt) = tmp; } while(0)
```
JonatanAntoni: After clarification with compiler experts I would add volatile here as well, just to be sure the actual register access is not optimized.

ithinuel: PR updated !

PR #394, opened 2018-07-13
Adding basic conv for nonsqaure and nn_mult functions

liangzhen-lai:


JonatanAntoni:
I recognized that you added new files to CMSIS-NN. Can you please check if those files needs to be included in the [NN Lib component](https://github.com/ARM-software/CMSIS_5/blob/b9918492389480f849aaaa869678e242fa24ca4b/ARM.CMSIS.pdsc#L2607).

PR #408, opened 2018-08-20
Removed ITM_TCR_GTSFREQ definition

Kochise:
See http://infocenter.arm.com/help/topic/com.arm.doc.ddi0337e/ch11s06s01.html#BABHACCJ (for Cortex-M3, but not present in official documentation)


JonatanAntoni:
Hi Kochise,

Thanks for revising the core headers.
You reference Cortex-M3 but change implementation for Cortex-M4. Is this by intend?
Can we align your fixes for all Cortex-Ms?

Cheers,
Jonatan

Kochise:
Yeah, the problem is that I have no much idea if that change covers all "Cortex-Ms" so I limited it to what I could cross check using the doc linked above.

JonatanAntoni:

PR #410, opened 2018-08-22
Fix define around EvrRtxKernelInitialized

c1728p9:
Rename the define EVR_RTX_KERNEL_INITIALIZE_COMPLETED_DISABLE to
EVR_RTX_KERNEL_INITIALIZED_DISABLE so it matches the reset of the
codebase.

The commit 858f429461e04921644f41eca9d7b63248434bd2 -
"RTX5: minor change in events (name consistency)" initially
changed this define but missed this one spot.


RobertRostohar:
Thanks for spotting and fixing.

Issue #415, opened 2018-08-25
IAR compilation problem with __RESTRICT define


c1728p9:
Building for IAR fails due to the `__restrict` keyword not being recognized. When `__RESTRICT` is defined to `restrict` builds work normally. This change was introduced in the commit 2fd08630a4ba98673ae5e2830a1689a59f99144a so 5.4.0 is the first release which contains this change.

For reference I found this issue when [upgrading mbed-os to version 5.4.0](https://github.com/ARMmbed/mbed-os/pull/7875).

TTornblom:
Looks like "__restrict" is only recognized in EWARM 8.x, while EWARM 7.x only recognizes "restrict", which needs the "--enable_restrict" compile option.

So this change needs rework to work with EWARM 7.

JonatanAntoni:
May I ask you both (@TTornblom and @c1728p9) to work out the solution together that works best for mbed? I don't have an EWARM7 testbench, here. A PR with a fix is very appreciated. 

Thanks,
Jonatan

c1728p9:
Closing this issue since #425 has been merged.

PR #425, opened 2018-09-04
IAR: IAR compilation problem with __RESTRICT define (7.80)

TTornblom:
Fix for #415

### DIFF
--- a/CMSIS/Core/Include/cmsis_iccarm.h
+++ b/CMSIS/Core/Include/cmsis_iccarm.h
@@ -150,7 +150,12 @@
 #endif
 
 #ifndef   __RESTRICT
-  #define __RESTRICT            __restrict
+  #if __ICCARM_V8
+    #define __RESTRICT            __restrict
+  #else
+    /* Needs IAR language extensions */
+    #define __RESTRICT            restrict
+  #endif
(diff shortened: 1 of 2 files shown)


JonatanAntoni:
@c1728p9 please double check this PR. I'll merge it asap after you confirmation.

CMSIS/Core/Include/cmsis_iccarm.h
```diff
@@ -150,7 +150,12 @@
 #endif
 
 #ifndef   __RESTRICT
-  #define __RESTRICT            __restrict
+  #if __ICCARM_V8
+    #define __RESTRICT            __restrict
+  #else
+    /* Needs IAR language extensions */
```
JonatanAntoni: @TTornblom is there a way to check for IAR language extensions? I guess the compiler creates weird error messages when using `restrict` without these extensions? Any chance to give a better hint to the user?

TTornblom:
Actually I tested this with mbed, which uses 7.80, and the default setting for 7.80 appears to be C99, where `restrict ` is supported without any need for extra compilation options.

The `--enable_restrict` option appears necessary only when compiling in C89 mode.

PR #430, opened 2018-09-11
DSP: fix double-promotion warnings in arm_mat_inverse_f64

andrewleech:
While using the DSP library in a project compiled with gcc and the `-Werror` setting I've  hit a number of `-Werror=double-promotion` instances in `arm_mat_inverse_f64.c`

This is triggered by comparing a wide float values with literals declared with the f suffix, eg `0.0f`

This patch resolves the issue by using the long float suffix instead.

For reference, this is the compiler error seen before this patch:
```
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c: In function 'arm_mat_inverse_f64':
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:211:16: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
       if (maxC == 0.0f)
                ^~
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:223:16: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
       if ( (in > 0.0f ? in : -in) != maxC)
                ^
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:236:23: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
           if ((*pInT2 > 0.0f ? *pInT2: -*pInT2) == maxC)
                       ^
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:283:31: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
       if ((flag != 1U) && (in == 0.0f))
                               ^~
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:672:29: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
     if ((flag != 1U) && (in == 0.0f))
                             ^~
cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c:677:20: error: implicit conversion from 'float' to 'float64_t {aka double}' to match other operand of binary expression [-Werror=double-promotion]
         if (pIn[i] != 0.0f)
                    ^~
cc1: all warnings being treated as errors
../../mkrules.mk:47: recipe for target 'build/cmsis/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.o' failed
```

### DIFF
--- a/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c
+++ b/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f64.c
@@ -87,7 +87,7 @@ arm_status arm_mat_inverse_f64(
 
   /* Run the below code for Cortex-M4 and Cortex-M3 */
 
-  float64_t Xchg, in = 0.0f, in1;                /* Temporary input values  */
+  float64_t Xchg, in = 0.0, in1;                /* Temporary input values  */
   uint32_t i, rowCnt, flag = 0U, j, loopCnt, k, l;      /* loop counters */
   arm_status status;                             /* status of matrix inverse */
 
@@ -155,18 +155,18 @@ arm_status arm_mat_inverse_f64(
       j = numRows - rowCnt;
       while (j > 0U)
       {
-        *pOutT1++ = 0.0f;
+        *pOutT1++ = 0.0;
         j--;
       }
 
       /* Writing all ones in the diagonal of the destination matrix */
-      *pOutT1++ = 1.0f;
+      *pOutT1++ = 1.0;
 
       /* Writing all zeroes in upper triangle of the destination matrix */
       j = rowCnt - 1U;
       while (j > 0U)
       {
-        *pOutT1++ = 0.0f;
+        *pOutT1++ = 0.0;
         j--;
       }
 
@@ -208,7 +208,7 @@ arm_status arm_mat_inverse_f64(
       }
 
       /* Update the status if the matrix is singular */
-      if (maxC == 0.0f)
+      if (maxC == 0.0)
       {
         return ARM_MATH_SINGULAR;
       }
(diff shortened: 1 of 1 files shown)


JonatanAntoni:
Thanks for pointing this out. 

The complete function uses `float64_t` (aka `double`) as data type but `f` suffix for constants. I am not sure why this was implemented that way. We will check how this can be cleaned up consistently.

andrewleech:
Cheers, yeah it's a pretty simple oversight to make, if you're not paying close attentions to warnings. 

I'm not sure if the `l` suffix used in this PR is the _correct_ fix, but it works for me.

andrewleech:
Ah drats, the project I'm using (micropython) has `-fsingle-precision-constant` which turns the un-suffixed literals into floats. 

So yes, no suffix is correct by C but still fails for me.

```
-fsingle-precision-constant causes floating-point constants to be loaded in single precision even when this is not exact. This avoids promoting operations on single precision variables to double precision like in x + 1.0/3.0. Note that this also uses single precision constants in operations on double precision variables. This can improve performance due to less memory traffic.
```

andrewleech:
Ok I've removed all instances of literal suffixes. 

Not yet sure how I'm going to fix my usage, but that's a different issue.

ps. the _f32 version of this file is close to identical, with all the same `f` suffixes, I suspect this file was originally made by copying that one.

andrewleech:
As far as I know this should be good to go. Is there anything else blocking this from being accepted?

I was able to fix my micropython project's usage of this by removing `-fsingle-precision-constant` at which point my changes here satisfy the build warnings correctly.

PR #446, opened 2018-10-05
IAR: Restructured NN examples

TTornblom:


ARM.CMSIS.pdsc
```diff
@@ -3840,11 +3840,11 @@ and 8-bit Java bytecodes in Jazelle state.
       </attributes>
     </example>
 
-    <example name="NN Library CIFAR10" doc="readme_iar.txt" folder="CMSIS/NN/Examples/IAR/iar_nn_examples/cifar10">
+    <example name="cifar10" doc="readme_iar.txt" folder="CMSIS/NN/Examples/IAR/iar_nn_examples/cifar10">
```
JonatanAntoni: I think the example name should be slightly more specific and expressive.

ARM.CMSIS.pdsc
```diff
@@ -3870,11 +3870,11 @@ and 8-bit Java bytecodes in Jazelle state.
       </attributes>
     </example>
 
-    <example name="NN Library GRU" doc="readme_iar.txt" folder="CMSIS/NN/Examples/IAR/iar_nn_examples/gru">
+    <example name="gru" doc="readme_iar.txt" folder="CMSIS/NN/Examples/IAR/iar_nn_examples/gru">
```
JonatanAntoni: I think the example name should be slightly more specific and expressive.  The name is used to identify the example project in pack explorer.

PR #473, opened 2018-11-30
CMSIS-Core(A): Fixed the position of "extern "C""

d-kato:
Fixed a bug that compilation fails if core_ca.h is included twice.


JonatanAntoni:
Thanks @d-kato for pointing this out.

PR #478, opened 2018-12-13
CM3: Make ACTLR bit definitions conditional

kjbracey-arm:
The ACTLR register itself is conditional on chip revision, but its bit definitions were always defined.

Make the the bit definitions also conditional, so it is possible to produce portable code that sets DISDEFWBUF if available:

    #ifdef SCnSCB_ACTLR_DISDEFWBUF_Msk
       SCnSCB->ACTLR |= SCnSCB_ACTLR_DISDEFWBUF_Msk;
    #endif


kjbracey-arm:
Alternative would be to make ACTLR not conditional on chip revision - presumably writing to it does nothing on early M3s.

But this style is consistent with the rest of the headers.

kjbracey-arm:
Thanks!

PR #485, opened 2018-12-20
Avoid __builtin_clz(0)

### LINKED ISSUE #463: Definition of __CLZ for gcc causes undefined behavior (formally)
For gcc `__CLZ` is defined in `cmsis_gcc.h` as follows:
``` cpp
/**
  \brief   Count leading zeros
  \details Counts the number of leading zeros of a data value.
  \param [in]  value  Value to count the leading zeros
  \return             number of leading zeros in value
 */
#define __CLZ             (uint8_t)__builtin_clz
```

According to [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html):

> Built-in Function: int __builtin_clz (unsigned int x)
> Returns the number of leading 0-bits in x, starting at the most significant bit position. **If x is 0, the result is undefined.**

(emphasis mine)

So, strictly speaking, __builtin_clz is not guaranteed to be compiled to CLZ instruction (or is it?) and may cause UB.

I suppose that's not very important but I still would prefer to see inline assembly instead of that builtin function.

kjbracey-arm:
`__builtin_clz(0)` is specified as undefined behaviour, so ensure `__CLZ` avoids it. Current ARM compilers eliminate the test for 0 at low optimisation levels, leaving just the CLZ instruction.

Resolves #463.

### DIFF
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -910,7 +910,23 @@ __STATIC_FORCEINLINE uint32_t __ROR(uint32_t op1, uint32_t op2)
   \param [in]  value  Value to count the leading zeros
   \return             number of leading zeros in value
  */
-#define __CLZ             (uint8_t)__builtin_clz
+__STATIC_FORCEINLINE uint8_t __CLZ(uint32_t value)
+{
+  /* Even though __builtin_clz produces a CLZ instruction on ARM, formally
+     __builtin_clz(0) is undefined behaviour, so handle this case specially.
+     This guarantees ARM-compatible results if happening to compile on a non-ARM
+     target, and ensures the compiler doesn't decide to activate any
+     optimisations using the logic "value was passed to __builtin_clz, so it
+     is non-zero".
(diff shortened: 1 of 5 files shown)


kjbracey-arm:
Dug further on Compiler Explorer: GCC 6.4 does not optimise this away; GCC 7.2 does.

JonatanAntoni:
Hi @kjbracey-arm,

thank you for providing this. I'd rather expect a change limited to compiler_gcc.h. Why do you think we should change armcc and armclang as well?

Cheers,
Jonatan

kjbracey-arm:
I'm not changing armcc.

clang does not have its own specification of `__builtin_clz` that I can find - it just seems to be being "gcc-compatible". In which case I think we have to assume `__builtin_clz` remains undefined. 

And presumably in principle cmsis_armclang.h could be renamed as `cmsis_clang.h` used for non-ARM targets just as `cmsis_gcc.h` is, according #463, even if not currently functional.

JonatanAntoni:
Hi @kjbracey-arm,

Ah, sorry. I'll crosscheck with some clang expert about the built-in in armclang.

Thanks again,
Jonatan

PR #486, opened 2018-12-21
RTOS2: IAR, change from INCLUDE to #include

TTornblom:


JonatanAntoni:
Hi @TTornblom,

Can you remove the binary libraries from your pull request, please?

We are currently discussing how to handle the pre-built libraries in the future. Updating them after each and every change to the source code would lead to a rapidly growing Git repository. So you are forces to copy *all* versions on every clone. That's getting a major issue. Hence we decided not to keep them up-to-date on the develop branch. 

Cheers,
Jonatan

TTornblom:
Removed the binary libraries commit.

I agree that shipping the pre-built libraries seems like a waste of space and bandwidth.

PR #487, opened 2018-12-21
RTX: make API more typesafe

ensc:
Actual API is very type unsafe because underlying opaque types are
simple 'void *' pointers and nearly everything can be casted implicitly
to it.

E.g.

| int a;
|
| osThreadFlagsSet(&a, 1);

will be accepted by C compilers with strictest warnings levels.

Patch makes 'osXXX_t' strongly typed so that such errors can be
detected at runtime.

Signed-off-by: Enrico Scholz <enrico.scholz@sigma-chemnitz.de>

### DIFF
--- a/CMSIS/RTOS2/Include/cmsis_os2.h
+++ b/CMSIS/RTOS2/Include/cmsis_os2.h
@@ -208,25 +208,32 @@ typedef enum {
  
  
 /// \details Thread ID identifies the thread.
-typedef void *osThreadId_t;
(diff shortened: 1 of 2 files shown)


JonatanAntoni:
Hi @ensc,

Thanks for your contribution. Unfortunately this type of strictness does not work in this place. The header file cmsis_os2.h is intended to be implementation agnostic. Hence there must not be any specific types from RTX5 implementation used. 

Cheers,
Jonatan

PR #496, opened 2019-01-07
RTOS/RTOS2: Updated IAR support

TTornblom:


ARM.CMSIS.pdsc
```diff
@@ -3224,6 +3224,8 @@ and 8-bit Java bytecodes in Jazelle state.
         <file category="library" condition="CM4_LE_GCC"      name="CMSIS/RTOS/RTX/LIB/GCC/libRTX_CM3_IFX.a" src="CMSIS/RTOS/RTX/SRC/GCC"/>
         <file category="library" condition="CM4_FP_LE_GCC"   name="CMSIS/RTOS/RTX/LIB/GCC/libRTX_CM4_IFX.a" src="CMSIS/RTOS/RTX/SRC/GCC"/>
         <!-- IAR -->
+        <file category="library" condition="CM4_LE_IAR"      name="CMSIS/RTOS/RTX/LIB/IAR/RTX_CM3_IFX.a" src="CMSIS/RTOS/RTX/SRC/IAR"/>
+        <file category="library" condition="CM4_FP_LE_IAR"   name="CMSIS/RTOS/RTX/LIB/IAR/RTX_CM4_IFX.a" src="CMSIS/RTOS/RTX/SRC/IAR"/>
```
JonatanAntoni: We never provided these libraries so far. RTX4 is deprecated. Why do you think its worth to add these?

CMSIS/RTOS/RTX/SRC/rt_CMSIS.c
```diff
@@ -269,7 +269,7 @@ static inline  t __##f (t1 a1, t2 a2, t3 a3, t4 a4) {                          \
 
 #elif defined (__ICCARM__)      /* IAR Compiler */
 
-#define __NO_RETURN __noreturn
+#define __NO_RETURN __attribute__((__noreturn__))
```
JonatanAntoni: Is this change backward compatible with all versions of IAR compiler? Or would this break anything?
RTX4 is deprecated. Why do you think its worth maintaining it? Any customer demands?

ARM.CMSIS.pdsc
JonatanAntoni: Its not about the library per se. Its about RTX4 at a whole. We do not maintain RTX4 any more. Instead we strongly recommend customers still using RTX4 to move to RTX5.

PR #497, opened 2019-01-07
RTOS2: Updated IAR support

TTornblom:


JonatanAntoni:
Hi @TTornblom,

Unfortunately enhancing the PDSC file but don't providing any pre-built libraries breaks the pack generation. 

Creating the Pack I get:
> *** ERROR M323: ARM.CMSIS.pdsc (Line 3321)
> File/Path not found: 'CMSIS\RTOS2\RTX\Library\IAR\RTX_V8MB.a'

Hence we need to provide initial versions for these libraries, now.
I'd like to take this opportunity to give Git LFS a try. I created features/lfs from right after this PR. I added a dummy RTX_V8MB.a which is correctly reported to be stored on LFS.

May I ask you to give this a try as well, please? You should be able to add all the libraries now (to that branch for the first try) and they should go to LFS instead of the repo itself.

Please let me know about your opinion.

Thanks,
Jonatan

TTornblom:
Hi @JonatanAntoni ,

I added the libraries to this commit.

Please check if I got it right.

Cheers,
Thomas

JonatanAntoni:
HI @TTornblom,

sorry, I guess my initial post was a but imprecise. I just updated it with correct links, see above.

You should be able to fetch the feature branch. Without pulling from LFS all the files should be "empty", i.e. contain only the hash of the file to be fetched from LFS. Actually pulling files from LFS should replace the libs with the binaries. 

Now you can add all the binaries and push them back. They should go to LFS. Inspecting the commit on GitHub should reveal this.

Please drop me the link to your fork of that branch or directly raise the PR if you think its works.

Cheers,
Jonatan

PR #500, opened 2019-01-08
Remove unnecessary inclusion of RTE_Components.h in RTX header

adeaarm:
One of the RTOS2 RTX headers, rtx_evr.h, includes directly the header RTE_Components.h, which seems to be not necessary as it is already included by RTX_Config.h.

With this patch the direct inclusion of the RTE_Components.h in rtx_evr.h is removed, rather than being put under the `#ifdef _RTE_`, as it seems to be redundant. If the inclusion is necessary, I still advise that to be put at least under the `#ifdef _RTE_`, so that builds that don't provide a RTE_Components.h will not fail.

### DIFF
--- a/CMSIS/RTOS2/RTX/Include/rtx_evr.h
+++ b/CMSIS/RTOS2/RTX/Include/rtx_evr.h
@@ -38,8 +38,6 @@
 #define   OS_EVR_WAIT           OS_EVR_THREAD
 #endif
 
-#include "RTE_Components.h"
-
 #ifdef    RTE_Compiler_EventRecorder
 
 //lint -emacro((835,845),EventID) [MISRA Note 13]


RobertRostohar:
Header rtx_evr.h is included also directly without previously including RTX_Config.h (for example in rtx_evr.c). Therefore it is necessary to include also RTE_Components.h.

Inclusion of RTE_Components.h could be guarded in rtx_evr.h with `#ifdef _RTE_` however this does not make a lot of sense since RTE_Components.h is still required.

RTE_Components.h needs to exist and define at least the CMSIS device header (CMSIS_device_header).

adeaarm:
rtx_evr.h always includes RTX_Config.h before this point (on line 30) which in turn includes RTE_Components.h, so by the time RTE_Components.h is directly included by rtx_evr.h, it would have already been included through RTX_Config.h, in every build that is going to provide an RTE_Components.h (i.e. that defines _RTE_), so this second inclusion would be always redundant in cases where there is an RTE_Components.h to include.

Our use case is different, we link with one of the pre-built RTX library packages available in CMSIS/RTOS2/RTX/Library/ . To successfully do so, we compile two source files, rtx_lib.c and RTX_Config.c . In particular rtx_lib.c includes rtx_evr.h , but we don't provide any RTE_Components.h as we don't need any particular symbol defined. With this change in place we wouldn't be able to build anymore even though we don't need to define any particular symbol in RTE_Components.h (as a workaround we should provide then an empty RTE_Components.h).

adeaarm:
Thanks for acknowledging. Yes, guarding the inclusion with `#ifdef _RTE_` is a solution that would work for us, please update the code accordingly.

RobertRostohar:
Please verify and close the issue if solved.

adeaarm:
Thanks, the proposed commit will work, closing this pull request.

Issue #499, opened 2019-01-08
Check asm volatile flagging - mostly excessive, but some may be missing


kjbracey-arm:
Scanning the codebase, I can see a load of clearly unnecessary `asm volatile`s - `__REVSH`, `__RBIT`, all `__UXTAB16` et al.

All of these are instructions with no side-effects, so can be treated like normal code, subject to elimination if their results are not accessed or the object they manipulate is not visible - this would be consistent with @JonatanAntoni's comment here: https://github.com/ARM-software/CMSIS_5/issues/463#issuecomment-448926726

Some like `__SSUB16` are justifiably volatile if we want to access the Q or GE flag side-effects, but then there seem to be some missing `volatile` like `__SSAT16`.

`__LDA` and `__STL` arguably have no need for `volatile`, once the missing barrier from #494 is present,  but taking off the `volatile` might put you at the cutting edge of optimisation there: https://stackoverflow.com/questions/45960387/why-dont-compilers-merge-redundant-stdatomic-writes

However, `__LDA` and `__LDRT` are all declared with volatile pointers as parameters, which rather dictates that they must do the access, which mandates a `volatile` on the assembler.

In C++11, there are volatile and non-volatile overloads for `atomic_load`, and CMSIS could do the same for its load/store intrinsics:

```
__STATIC_FORCEINLINE uint32_t __LDRT(volatile uint32_t *ptr)
{
    uint32_t result;

   __ASM volatile ("ldrt %0, %1" : "=r" (result) : "Q" (*ptr) );
   return(result);
}

#ifdef __cplusplus
__STATIC_FORCEINLINE uint32_t __LDRT(uint32_t *ptr)
{
    uint32_t result;

   __ASM ("ldrt %0, %1" : "=r" (result) : "Q" (*ptr) );
   return(result);
}
#endif
```
Clearly this is low-priority, as there's nothing actually "wrong" except maybe `SSAT16` - the rest are just maybe suboptimal.

JonatanAntoni:
Hi @kjbracey-arm,

I concluded on your PR's and merged them all. Can you please cross check if we hit everything?

Thanks,
Jonatan

kjbracey-arm:
Closed all the PRs - I'm not aware of any remaining issues.

Make sure to let me know if any problems occur due to those changes.

PR #501, opened 2019-01-11
GCC: remove unneeded asm volatiles

kjbracey-arm:
Many assembler intrinsics are pure - they read only the input registers, write only the output registers, and change no PSR flags.

Remove volatile from all such intrinisics to improve potential optimisation.

   REV, REV16, REVSH, RBIT
   QADD8, SHADD8, UQADD8, UHADD8
   QSUB8, SHSUB8, UQSUB8, UHSUB8
   QADD16, SHADD16, UQADD16, UHADD16
   QSUB16, SHSUB16, UQSUB16, UHSUB16
   QASX, SHASX, UQASX, UHASX
   QSAX, SHSAX, UQSAX, UHSAX
   USAD8, USADA8
   UXTB16, UXTAB16, SXTB16, SXTAB16
   SMMLA

Relates to issue #499.

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -904,7 +904,7 @@ __STATIC_FORCEINLINE uint32_t __REV(uint32_t value)
 #else
   uint32_t result;
 
-  __ASM volatile ("rev %0, %1" : __CMSIS_GCC_OUT_REG (result) : __CMSIS_GCC_USE_REG (value) );
+  __ASM ("rev %0, %1" : __CMSIS_GCC_OUT_REG (result) : __CMSIS_GCC_USE_REG (value) );
   return result;
 #endif
 }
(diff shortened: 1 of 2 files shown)


kjbracey-arm:
Just added a second commit fixing up the odd declaration of REV16 in Core_A version.

grubbymits:
SADD16 writes to the GE bits, whereas QADD16 doesn't read or write to any bits in the APSR.

GuentherMartin:
I see.
Does this mean that 'asm volatile' is necessary if any flags are updated?

sparker-arm:
Yes, it will be used to prevent instructions from being reordered where data dependencies aren't obvious to the compiler. For sadd16, we wouldn't want it to be moved past the sel instruction that reads the GE flags.

GuentherMartin:
So 'asm volatile' is used not because of the flags but because of the possible reorder of instructions.
In other words: 'asm volatile' is used when compiler optimization is not wanted/allowed.

GuentherMartin:
Thank you for the explanation.

kjbracey-arm:
FWIW, you could also force ordering between PSR-setting and PSR-using ops by using "memory" clobbers ("this reads or writes unspecified memory") on all PSR-related things, but that almost certainly inhibits more wanted optimisations than "volatile" - it forces a register spill to memory for whatever data you were manipulating, and really gums up scheduling.

And using "cc" clobbers doesn't work - I tried it. GCC only interprets that as "writes to flags", not "reads from flags", so you can't use it to keep a MRS or SSEL after a SADD8. (The "memory" clobber is special, considering the possibility of reads - all other clobbers are write-only).

JonatanAntoni:
Merged in efa5af1 and 4d8dc46

PR #502, opened 2019-01-11
IAR: LDRT et al must be asm volatile

kjbracey-arm:
As these functions take volatile pointers, the API is promising that the loads and stores will happen, so the assembler statements need volatile qualifiers too.

If the functions took non-volatile pointers, or had a separate non-volatile overload for C++, then the volatile could be omitted - the instructions are normal loads and stores with no side-effects.

GCC and clang assembler already is "asm volatile", and armcc uses intrinsics.

Relates to issue #499.


JonatanAntoni:
Merged in a6f49fd

PR #503, opened 2019-01-11
GCC: SSAT and USAT need asm volatile

kjbracey-arm:
Both SSAT and USAT modify the Q flag, so need something to order them
with respect to reads of the PSR. Add `volatile`, matching SADD8 et al.

Also SSAT16 and USAT16.

Relates to issue #499.


kjbracey-arm:
CMSIS documentation does not state that USAT and SSAT set the Q flag, and their ARMv6 substitutes obviously don't, but USAT16 and SSAT16 do document it. It seems reasonable that USAT and SSAT should too when on appropriate hardware.

JonatanAntoni:
Merged in a4e6957

PR #504, opened 2019-01-11
GCC: add WFI/WFE compiler barriers

kjbracey-arm:
Add "memory" clobber to __WFI and __WFE. Architecturally these should always be immediately preceded by a __DSB (eg to ensure the write buffer is drained). Without a barrier on WFI, the following compiler reordering would be permitted:

     __DSB();             __DSB();
     __WFI();       ->    val_in_ram = 3;
     val_in_ram = 3;      __WFI();

This could cause some power issues with the external bus not being idle.

The added barrier should have no impact on code size, assuming these instructions are always accompanied by DSB, as DSB does have its own memory clobber already.

SEV not modified as there are no issues with the equivalent reordering; we only need the SEV to not be reordered before the DSB, which is ensured by volatile.

Relates to issue #499.

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -839,7 +839,7 @@ __STATIC_FORCEINLINE void __set_FPSCR(uint32_t fpscr)
   \brief   Wait For Interrupt
   \details Wait For Interrupt is a hint instruction that suspends execution until one of a number of events occurs.
  */
-#define __WFI()                             __ASM volatile ("wfi")
+#define __WFI()                             __ASM volatile ("wfi":::"memory")
 
 
 /**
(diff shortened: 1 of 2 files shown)


kjbracey-arm:
Read a bit deeper on this - the DSB I talk about isn't *architecturally required*, just usually a good idea. Two reasons:

1) It ensures any writes that anyone else is waiting for reach memory, so we don't stall the system cos they're stuck in our write buffer while we sleep.
2) Having our memory interface quiescent may save power during sleep.

However, the DSB often isn't actually there, eg in a typical WFE spinlock, because other ordering basically forces the issue on point 1 (eg the combination of a release barrier on a store followed by a load for the spinlock), and it wouldn't be effective anyway for point 2 if IRQs were enabled.

So this may actually inhibit some WFE optimisations in those cases where people haven't been bothering with the DSB, but it's generally better safe than sorry with these instructions, if using assembler - proper optimisation would need use of compiler built-ins anyway.

JonatanAntoni:
Merged in a693d67

PR #507, opened 2019-01-14
GCC/Clang: cope with __set_CPSR changing condition codes

kjbracey-arm:
Add a "cc" clobber to `__set_CPSR` to reflect the fact that it could be changing the condition codes.


JonatanAntoni:
Merged in 4f14f49

Issue #493, opened 2019-01-03
NVIC_EnableIRQ missing barrier?


kjbracey-arm:
I've been getting paranoid about compiler barriers, particularly as people have been looking at Link-Time Optimisation.

`NVIC_DisableIRQ` has the DSB/ISB sequence needed by the architecture, and they act as the necessary "acquire" compiler barrier to keep following code in line - don't access RAM until that IRQ disable has completed.

But `NVIC_EnableIRQ` has nothing - it's just a `volatile` memory access. If code does something like:

    NVIC_DisableIRQ(devx);
    // modify some RAM accessed by devx IRQ handler
    NVIC_EnableIRQ(devx);

Nothing is ensuring that RAM access occurs before the IRQ is enabled. `volatile` doesn't say anything about ordering of non-volatile accesses. Eg gcc manual: 

> Accesses to non-volatile objects are not ordered with respect to volatile accesses. You cannot use a volatile object as a memory barrier to order a sequence of writes to non-volatile memory. 

If `NVIC_DisableIRQ` acts as a barrier, it makes sense that `NVIC_EnableIRQ`  should also have one, so that the pair can function as a "lock/unlock" like above.

So my suggestion is:

    _STATIC_INLINE void __NVIC_EnableIRQ(IRQn_Type IRQn)
    {
      if ((int32_t)(IRQn) >= 0)
      {
        __COMPILER_BARRIER();
        NVIC->ISER[(((uint32_t)IRQn) >> 5UL)] = (uint32_t)(1UL << (((uint32_t)IRQn) & 0x1FUL));
        __COMPILER_BARRIER(); // not sure if 2nd needed - better safe than sorry
      }
    }

Would need a barrier macro in the compiler-specific stuff.

kjbracey-arm:
I note that `__enable_irq()` and `__disable_irq()` also act as barriers, via their assembler memory clobber, so this would be consistent with those.

ReinhardKeil:
Thanks for your comments.  Please keep in mind, barrier instructions are not free.  They consume both memory and execution time.

For an interrupt enable, the barrier instruction is not a strict requirement, as a potential interrupt that is triggered by the code that follows will be served perhaps, just a few cycles later or earlier.  

For an interrupt disable, it is however critical that the interrupt is for sure disabled before you executed further instructions.

Did you discover a real-world problem that we have missed?  Do you have a code example that does not work due to the fact that there are not barrier instructions for NVIC_Enable?

[...]

PR #510, opened 2019-01-16
__NVIC_EnableIRQ compiler barriers 

kjbracey-arm:
__NVIC_DisableIRQ and __NVIC_EnableIRQ can be used to function as a mutex-style protection lock against a particular interrupt handler, similar to __disable_irq and __enable_irq for all interrupts.

However, __NVIC_EnableIRQ, unlike a mutex unlock or __enable_irq, had no compiler barriers. Being just a volatile write, in the following code sequence:

    NVIC_DisableIRQ(devx);
    // modify some RAM accessed by devx IRQ handler
    NVIC_EnableIRQ(devx);

there would be nothing preventing the RAM accesses from moved below the NVIC_EnableIRQ.

Add barriers to NVIC_EnableIRQ, so that the above code works the same as a mutex or __disable_irq, without any added need to mark the shared RAM as volatile.

Fixes issue #493

### DIFF
--- a/CMSIS/Core/Include/cmsis_armcc.h
+++ b/CMSIS/Core/Include/cmsis_armcc.h
@@ -104,6 +104,9 @@
 #ifndef   __RESTRICT
   #define __RESTRICT                             __restrict
 #endif
+#ifndef   __COMPILER_BARRIER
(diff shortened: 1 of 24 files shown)


0xc0170:
What is needed for this PR to progress?

PR #512, opened 2019-01-17
Fix issue #511.

### LINKED ISSUE #511: Bug in arm_maxpool_q7_HWC
When the following macro `ARM_MATH_DSP`

https://github.com/ARM-software/CMSIS_5/blob/362bc60946c5626b3f090249a181bed480c86e47/CMSIS/NN/Source/PoolingFunctions/arm_pool_q7_HWC.c#L163-L173

is enabled, it calls
https://github.com/ARM-software/CMSIS_5/blob/362bc60946c5626b3f090249a181bed480c86e47/CMSIS/NN/Source/PoolingFunctions/arm_pool_q7_HWC.c#L211

The problem is that `compare_and_replace_if_larger_q7` handles only input size of multiple of 4, see its implementation below

https://github.com/ARM-software/CMSIS_5/blob/362bc60946c5626b3f090249a181bed480c86e47/CMSIS/NN/Source/PoolingFunctions/arm_pool_q7_HWC.c#L52-L82

It does not handle the case when `length` is not multiple of 4.

csukuangfj:
Handle the case in max pooling when the number of channels is not multiple of 4.

See the description of #511.

Fix #511.

### DIFF
--- a/CMSIS/NN/Source/PoolingFunctions/arm_pool_q7_HWC.c
+++ b/CMSIS/NN/Source/PoolingFunctions/arm_pool_q7_HWC.c
@@ -50,12 +50,12 @@ static void buffer_scale_back_q15_to_q7(q15_t * buffer, q7_t * target, uint16_t 
 }
 
 static void compare_and_replace_if_larger_q7(q7_t * base,   // base data
-                                             q7_t * target, // compare target
+                                             const q7_t * target,   // compare target
                                              const uint16_t length  // data length
     )
 {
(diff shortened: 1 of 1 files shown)


felix-johnny:
@csukuangfj  .. Could you rebase this patch to get rid of the auto rebase issue? Looks good otherwise..

csukuangfj:
@felix-johnny 

Done.

felix-johnny:
@csukuangfj .. Thanks!

PR #516, opened 2019-01-17
Core(A)/armclang: Remove ISB/DSB/DMB barriers

kjbracey-arm:
Core(M) versions of files already do not have explicit barriers, so this makes Core(A) consistent - the built-ins are specified as having barriers anyway.


JonatanAntoni:
Merged as [#2d5351b](https://github.com/ARM-software/CMSIS_5/commit/2d5351b5db44210b3a669385671ee9e255783726)

PR #517, opened 2019-01-17
ARMCC: remove explicit DSB/DMB/ISB barriers

kjbracey-arm:
ARMCC documentation states that the `__dsb` etc intrinsics act as optimisation barriers. Even though that's a bit woolly about the exact equivalent barrier intrinsic, take its word that it's doing the right thing.

It seems safe to assume that it is, because the `__schedule_barrier()` intrinsics here are not actually sufficient for DSB and DMB. They need a `__memory_barrier()`. If no-one has seen any problems, then presumably
they already include one.


JonatanAntoni:
Merged in f88880f

PR #518, opened 2019-01-17
Revert "Core(M): Added memory clobbers to get_PRIMASK..."

kjbracey-arm:
GCC bug was fixed in October 2017 - change went to GCC 6.5, 7.3 and 8.1.

Remove the unnecessary memory clobbers to improve optimisation, on the assumption that people will have updated their compilers.

This reverts commit 35205629798192794d4918cef1d949141e32de4e.


JonatanAntoni:
Merged in 352e0ac

kjbracey-arm:
Might be worth putting info about compiler versions in a release note (in case anyone reads such things), or updating "requirements" documentation to cover this known bug.

PR #519, opened 2019-01-17
GCC: ISB does not require a memory clobber

kjbracey-arm:
ISB purely synchronises the instruction stream - there is no inherent need to include a memory clobber.

If memory accesses need to be synchronised, there would be some other neighbouring intrinsic such as DMB with a memory clobber, and this would synchronise against that via the volatile.

In other cases, this might be synchronising against just an MCR or similar, so avoiding the memory clobber in that case can improve optimisation - again ordering will be maintained via the volatile.

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -865,7 +865,7 @@ __STATIC_FORCEINLINE void __set_FPSCR(uint32_t fpscr)
  */
 __STATIC_FORCEINLINE void __ISB(void)
 {
-  __ASM volatile ("isb 0xF":::"memory");
+  __ASM volatile ("isb 0xF");
 }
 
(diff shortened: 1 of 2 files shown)


kjbracey-arm:
I've been pondering it overnight, and have decided this isn't safe.

ISB does too much work, and we don't really know what it's there for. It may not require a memory clobber, but we want some ordering constraint to try to keep following code after it.

And in GCC the only ordering constraint we have versus normal code is the memory clobber. Theoretical example: 

         write to CPACR to enable FPU
         ISB to make sure it takes effect
         Do some FP operations in C

There's no barrier in GCC that could totally stop any FP operation moving ahead of the ISB, or indeed the CPACR change, but in practice a memory clobber will usually work, by stopping the FP operands being loaded from memory before the ISB.

So I think I'll close this.

PR #521, opened 2019-01-18
Core(M): Add LDA/STL memory clobbers

### LINKED ISSUE #494: __LDA(EX)/__STL(EX) don't constrain compiler reordering
CMSIS has `__LDA`/`__STL` and `__LDAEX`/`__STLEX` family of functions, but the toolchains are inconsistent about what compiler barriers are present. Some functions use `volatile asm` with no compiler barriers, some use `volatile asm` with a memory clobber (ie a barrier), others use compiler built-ins, which will have compiler acquire/release semantics.

As some definitions lack any barrier, safe use of these macros requires code like:

    __LDAEX/STLEX for atomic_flag_test_and_set
    __compiler_barrier to stop following code moving ahead of __LDAEX
    do work protected by flag
    __compiler_barrier to stop preceding code moving after __STL
    __STL for atomic_flag_clear

But that's excessively pessimistic if the functions are using built-ins - adding the barriers loses some of the benefit of the builtins - having less-restrictive ordering than a complete fence.

It would be better if the assembler forms met the minimum expected acquire/release compiler semantics - I believe that can only be done as a full fence via a "memory" clobber. But having that full fence as a clobber only in the assembler versions, rather than all the time for every platform in the user's code is preferable.

Then user code can safely be:

    __LDAEX/STLEX for atomic_flag_test_and_set
    do work protected by flag
    __STL

I note that more functions could be using built-ins, avoiding the need for a memory clobber - `__LDA` and `__STL` could be implemented using `__c11_atomic_load` (clang) or `__atomic_load_4` (gcc).  (Assuming they map - do they always if ARMv8, or is there a "SMP" compiler flag that activates CPU-ordering instructions inside atomics?)

kjbracey-arm:
Constrain compiler reordering around acquire and release accesses by adding memory clobbers to assembler fragments for LDA/STL and LDAEX/STLEX.

Fixes #494.

### DIFF
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -1195,7 +1195,7 @@ __STATIC_FORCEINLINE uint8_t __LDAB(volatile uint8_t *ptr)
 {
   uint32_t result;
 
-  __ASM volatile ("ldab %0, %1" : "=r" (result) : "Q" (*ptr) );
+  __ASM volatile ("ldab %0, %1" : "=r" (result) : "Q" (*ptr) : "memory" );
   return ((uint8_t) result);
 }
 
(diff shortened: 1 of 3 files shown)


JonatanAntoni:
Merged in baf1da4d

Issue #405, opened 2018-08-16
Startup init order


KarolisMil:
In CMSIS/Device folder, GCC toolchain files:
why SystemInit is called after static data initialization (copying, setting) and not before? Every argument that comes to my mind says that this can be improved by calling SystemInit at the very start. I would love to know the reasons why it is like it is. Thank you!

ReinhardKeil:
I agree this is strange.  The issue might be that clearing memory also deletes stack content. We need to verify that.

KarolisMil:
After SystemInit, system could be more versatile. One of many examples migth be with clocks configured in SystemInit. That means data copy and setting ZI might be done faster if so. But this is just as example and it is up to user what to do there in SystemInit. The question is wheather he/she has that oportunity or needs to change official example files before use...

salkinium:
`SystemInit` may be written in C and may therefore use all of C’s features like reading a initialized `.data` variable in SRAM. If `SystemInit` is called before the `.data` or `.bss` sections have been copied/initialized, you may read random, or after a software restart, even old values.
So unless you write `SystemInit` in Assembly, this is a bit of a 🐔 and 🥚 problem.

ilg-ul:
@salkinium, this sounds like a bit of a 🐔 and 🥚 problem, but you can also write `SystemInit()` in assembly and, if careless, access external variables that were not initialised; the same as you can write it in C and, if careful, avoid such variables. The language you use to write this function is irelevant, only it's content is.

However it is true that calling `SystemInit()` before `.data` or `.bss` sections have been copied/initialized has the disadvantage that it cannot leave any results in global variables, because they will be erased shortly.

Actually, for Cortex-M, the entire startup sequence can be written in C, the only real danger is that without some cryptic options, modern compilers may not like pointers initialised from linker scripts, and trigger undefined behaviour.

salkinium:
> the same as you can write it in C and, if careful, avoid such variables.
> Actually, for Cortex-M, the entire startup sequence can be written in C

Yeah, at least for GCC that works fine, since it behaves like you'd expect.
Don't ask a compiler engineer about this though, they'll tell you all about the C Abstract Machine and UB until your ears fall off. That's why all the startup scripts are still in asm even though it's marketed as "C compatible". #AtLeastTheStackPointerIsSetInHardware

> without some cryptic options, modern compilers may not like pointers initialised from linker scripts

[...]

PR #524, opened 2019-01-23
Call SystemInit in the very beginning of Reset_Handler

alfedotov:
Related issue #405

### DIFF
--- a/CMSIS/DoxyGen/Core/src/Ref_SystemAndClock.txt
+++ b/CMSIS/DoxyGen/Core/src/Ref_SystemAndClock.txt
@@ -18,8 +18,8 @@ a more flexible configuration of the microcontroller system.
 used throughout the whole system initialization and runtime to calculate frequency/time related values.
 Thus one must assure that the variable always reflects the actual system clock speed. Be aware that
 a value stored to \c SystemCoreClock during low level initializaton (i.e. \c SystemInit()) might get
-overwritten by C libray startup code. Thus its highly recommended to call \ref SystemCoreClockUpdate
-at the beginning of the user \c main() routine.
+overwritten by C libray startup code and/or .bss section initialization.
+Thus its highly recommended to call \ref SystemCoreClockUpdate at the beginning of the user \c main() routine.
 
 
 \section system_init_code_ex_sec Code Example     
--- a/Device/ARM/ARMCM0/Source/GCC/startup_ARMCM0.S
+++ b/Device/ARM/ARMCM0/Source/GCC/startup_ARMCM0.S
@@ -119,6 +119,8 @@ __Vectors_End:
                 .globl   Reset_Handler
                 .fnstart
 Reset_Handler:
+                bl       SystemInit
+
 /* Firstly it copies data from read only memory to RAM.
  * There are two schemes to copy. One can copy more than one sections.
  * Another can copy only one section. The former scheme needs more
@@ -246,7 +248,6 @@ Reset_Handler:
 .L_loop3_done:
 #endif /* __STARTUP_CLEAR_BSS_MULTIPLE || __STARTUP_CLEAR_BSS */
 
-                bl       SystemInit
                 bl       _start
 
                 .fnend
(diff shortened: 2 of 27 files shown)


ilg-ul:
Could you also update the documentation to state that calling `SystemCoreClockUpdate()` is mandatory after the .bss section is cleared?

alfedotov:
> 
> 
> Could you also update the documentation to state that calling `SystemCoreClockUpdate()` is mandatory after the .bss section is cleared?
That's interesting thing because .bss can be cleared by library code (newlib for instance), not only cmsis startup code.

alfedotov:
> The ideal solution, that I used in my projects, is to have two system initialisation routines, one called as early as possible (to initialise PLL & external RAM, for example; it can be called `SystemEarlyInit()`), and the second called after the .data & .bss sections are initialised (for all other things, including clock variable initialisation; it can be the current `SystemInit()`), but before the C++ static constructors are executed.

This would be nice solution I agree.
I'm thinking to refactor newlib startup code (at the moment it is a complete hell of ifdef/else vs rdimon/bare-config blocks...) and probably we could tie them more elegantly.

ilg-ul:
> refactor newlib startup code

I don't think that it is reasonable to expect newlib to provide the startup code for CMSIS applications, newlib is a generic library, it may or may not be used by an application.

Not to mention that other toolchains may have different startup requirements.

The startup code must match the linker script and the memory allocator, and, from my experience, the only reasonable place where you can control all three is in the application.

JonatanAntoni:
Hi all,

please bear in mind that current convention is that `SystemInit()` is exactly intended to what @ilg-ul  wants to use it, e.g. early stage external memory configuration. It should run before any library initialization which might otherwise access non-configured external memory. Actually it does exactly this for all compilers other than GCC. Hence I think we should definitively fix it this way.

Having a second hook somewhere in between is another story, e.g. after `.bss` init but before C++ static constructors. I am not an expert on that field. Currently CMSIS recommendation is to call `SystemCoreClockUpdate()` right at the beginning of `main()`. If you think that's too late please feel free to suggest enhancements you see valuable. Don't limit to GCC but take other compilers and libraries into account as well. At which point do you want to see an additional standardized CMSIS hook? How to actually inject it into Arm Compiler, IAR *and* GCC startup?

Thanks for all your contributions.

Cheers,
Jonatan

ilg-ul:
> `SystemCoreClockUpdate()` right at the beginning of `main()`

this might be ok for C, but it is too late for C++.

for GCC, the right place for a second set of system inits is in the startup code, which must be placed in the application, not in the system libraries.

for other unfortunate toolchains... I don't know, but I guess it must be possible to disable the default startup code (same as `-nostartfiles` does for GCC), and allow the startup to be in the application, and be customised as needed.

ilg-ul:
anyway, if I understood it right, the changes apply to the sample files.

vendors will continue to publish their own startup files, where we have no control where `SystemInit()` is placed.

ilg-ul:
> Well I like that approach with SystemEarlyInit() and it perfectly fits with thing I want to get eventually.

for a more elaborate startup, you can take a look at:

https://github.com/micro-os-plus/startup-xpack/blob/xpack/src/startup.cpp

it is intended to be portable, I already used it in some RISC-V projects and I plan to use it for Cortex-M in the next release of µOS++.

JonatanAntoni:
@alfedotov, no we can't. `SystemInit()` is currently defined to be called at early stage. It is exactly implemented this way for other compilers than GCC. So switching between compilers would lead to different behavior which is not the intention of CMSIS.

@ilg-ul, that's the bad part of it. The startup code discussed here is only a template for the templates shipped by silicon vendors as part of their device family packs. Finally the startup code is under full control of the application, intended to be adopted to users needs. That's why fixing it here is not urgent I think.

ilg-ul:
> fixing it here is not urgent I think.

right, it is not urgent, but I think that as template, which apparently does no harm to existing applications, it should probably show the call to `SystemInit()` as early as possible.

and I also think you should really rework the documentation and explain at large why `SystemInit()` should be called early and the need to call `SystemCoreClockUpdate()` later, mandatory right before the static constructors for C++.

PR #529, opened 2019-02-07
RTX5: Add CLREX to exception exit paths

kjbracey-arm:
Safe use of LDREX/STREX requires that any thread switches, interrupts or
any other disruption between the LDREX and STREX be visible, so
exception handlers must ensure that the exclusive monitor flag is clear
when returning to foreground code.

Two possible error cases are

~~~~
   <thread 1>
   LDREX          exclusive monitor set
   <thread switch 2>
   STR            exclusive monitor cleared
   LDREX          exclusive monitor set
   <thread switch 1>
   STREX          !!! store succeeds because exclusive monitor set
~~~~
or
~~~~
   <thread>
   LDREX          exclusive monitor set
   <IRQ>
   STR            exclusive monitor cleared
   LDREX          exclusive monitor set
   // No store because of failed compare-and-swap
   <IRQ return>
   STREX          !!! store succeeds because exclusive monitor set
~~~~
For ARMv7-M and ARMv8-A, the monitor is automatically cleared on
exception return, but ARMv7-A requires that the clear be performed in
software.

Unneeded CLREX removed from data abort entry.  The entry state of the
exclusive monitor may be unknown on entry to an exception handler, but
no code can possibly be relying on it being clear - an STREX that is
intended to store must be preceded by an LDREX.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/ARM/irq_ca.s
+++ b/CMSIS/RTOS2/RTX/Source/ARM/irq_ca.s
@@ -100,6 +100,7 @@ Undef_Cont
                 SUB     LR, LR, R0
                 LDR     R0, [SP, #28]               ; Restore stacked SPSR
                 MSR     SPSR_CXSF, R0
+                CLREX                               ; Clear exclusive monitor
                 POP     {R0-R4, R12}                ; Restore stacked APCS registers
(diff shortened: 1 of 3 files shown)


JonatanAntoni:
Phew. Thanks @kjbracey-arm for pointing that out.

VladimirUmek:
All ok, thanks for this pull.

PR #531, opened 2019-02-11
Const-qualify all read-only pointers in CMSIS-DSP (#349)

### LINKED ISSUE #349: Constness of input parameters for CMSIS-DSP functions
Some functions of CMSIS-DSP library take vectors as input parameters (passed by pointer to the first element).
Some of these functions don't modify at all these vectors and so it would be better if the type of the input parameters is `const _type_* pSrc`.
As an example `arm_max_xxx()`, `arm_mean_xxx()`, `arm_min_xxx()`, and so on.
For other functions it's not clear if they can work in place by design (and so input vector cannot be pointer-to-const) or not (and so input vector could be declared as pointer-to-const).
As an example,I refer to `arm_q31_to_float()` and so on.

Could you review the declarations of the functions in CMSIS-DSP library and set the constness when needed?

torfinnberset:
This PR closes #349. The lack of const-correctness has been bugging me lately as I've been writing a C++ wrapper around the CMSIS-DSP library, and my C++ linter and compiler are quite strict in requiring const correctness. 

I've also fixed a couple of recurring typos. Adding `const` might in some cases help the compiler optimize further as well, but this has not been verified. 

In some source files, a pointer has been reused to point to both a `const` and non-`const` field. In these cases, I've elected to do a cast from `const` to non-`const`, otherwise I'd be introducing a second temporary pointer.

### DIFF
--- a/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/abs.c
+++ b/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/abs.c
@@ -1,7 +1,7 @@
 #include "ref.h"
 
 void ref_abs_f32(
-  float32_t * pSrc,
+  const float32_t * pSrc,
   float32_t * pDst,
   uint32_t blockSize)
 {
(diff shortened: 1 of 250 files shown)


torfinnberset:
(Tested on GCC 8.2.1 with -Werror)

PR #535, opened 2019-02-18
Revert "Const-qualify all read-only pointers in CMSIS-DSP (#349)"

JonatanAntoni:
Reverts ARM-software/CMSIS_5#531 due to compilation issues. @GuentherMartin will provide further details, soon.

cc: @torfinnberset, @llefaucheur, @MatthiasHertel80

### DIFF
--- a/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/abs.c
+++ b/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/abs.c
@@ -1,7 +1,7 @@
 #include "ref.h"
 
 void ref_abs_f32(
-  const float32_t * pSrc,
+  float32_t * pSrc,
   float32_t * pDst,
   uint32_t blockSize)
 {
--- a/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/add.c
+++ b/CMSIS/DSP/DSP_Lib_TestSuite/RefLibs/src/BasicMathFunctions/add.c
@@ -1,8 +1,8 @@
 #include "ref.h"
 
 void ref_add_f32(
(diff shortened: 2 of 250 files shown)


torfinnberset:
I think it would be greatly beneficial if there were some CI running for this repo, to catch these issues before merging

JonatanAntoni:
Hi @torfinnberset,

we have some sort of CI running here but unfortunately not for DSP, yet.
There is room for improvement. @MatthiasHertel80 is working on that.

Cheers,
Jonatan

GuentherMartin:
The original pull request (#349) causes several problems.
a) The reference libraries could not be build any longer.
b) casting lead to wrong usage of a pointer.
   a pointer is initialized with a pointer to const using a cast. Then later data is modified using this pointer which is wrong.
e.g.: see pointer pb in file arm_lms_f32.c line 210, 221

escherstair:
I suspected from the beginning that declaring the proper constness for the pointers would have been requiring a lot of work.
I agree that static analysis, CI and unit test is a "must have" for this repo.
Github offers some nice tools free of charge for open projects.

JonatanAntoni:
Hi @escherstair,

I agree with you that having some sort of CI running would be beneficial. But the fact that we are talking about embedded software makes things a little bit harder.

We could try to give it a start using free tools but that would need some community contribution. Are you familiar with the stuff GitHub offers? What would it take to set up some basic checks?

Cheers,
Jonatan

escherstair:
Hi @JonatanAntoni 
I'm not an expert, but I can help in some way.

First of all I need to understand what are your targets for the library:
- which compilers do you want to support? Which compiler/toolchain do you use now for your tests?
- are you interested in some kind of static code analysis?
- how do you perform unit testing at the moment?

JonatanAntoni:
@escherstair lets move this discussion of this thread to a dedicated discussion at #536.

Issue #550, opened 2019-03-13
v8-M: Contribution proposal for priority boosting functions, PRIS bit in SCB->AIRCR


MicroConsult:
I am missing functions to set/clear/get the PRIS bit in SCB->AIRCR.
Use case is the enabling/disabling of priority boosting of armv8-M.
It applies to CM33 and may be added to CM23 as well (after AICRC was added to the CM23.h).

Please find my proposal for an extension of the core_CM33.h header file below.

`#if defined (__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U)
#define NVIC_SetPriorityBoosting    __NVIC_SetPriorityBoosting
#define NVIC_ClearPriorityBoosting  __NVIC_ClearPriorityBoosting
#define NVIC_GetPriorityBoosting    __NVIC_GetPriorityBoosting

/**
  \brief   Set Priority Boosting
  \details Sets the priority boosting bit using the required unlock sequence.
 */
__STATIC_INLINE void __NVIC_SetPriorityBoosting(void)
{
  uint32_t reg_value;
    
  reg_value  =  SCB->AIRCR;                                     /* read old register configuration    */
  reg_value &= ~SCB_AIRCR_VECTKEY_Msk;                          /* clear bits to change               */
  reg_value  =  (reg_value                                   |
                ((uint32_t)0x5FAUL << SCB_AIRCR_VECTKEY_Pos) |
                SCB_AIRCR_PRIS_Msk                      );      /* Insert write key and priority boosting bit */
  SCB->AIRCR =  reg_value;
}

/**
  \brief   Clear Priority Boosting
  \details Clears the priority boosting bit using the required unlock sequence.
 */
__STATIC_INLINE void __NVIC_ClearPriorityBoosting(void)
{
  uint32_t reg_value;
    
  reg_value  =  SCB->AIRCR;                                     /* read old register configuration    */
  reg_value &= ~(SCB_AIRCR_VECTKEY_Msk | SCB_AIRCR_PRIS_Msk);   /* clear bits to change               */
  reg_value  =  (reg_value                                 |
                ((uint32_t)0x5FAUL << SCB_AIRCR_VECTKEY_Pos));  /* Insert write key                   */
  SCB->AIRCR =  reg_value;
}

/**
  \brief   Get Priority Boosting
  \details Reads the priority boosting bit from the NVIC Interrupt Controller.
  \return  0  Interrupt Priority Boosting is disabled
[...]

PR #551, opened 2019-03-14
armv8-M core_cm33.h add functions for secure interrupt priority boosting

MicroConsult:
Adding 4 new functions NVIC_Get/Set/Clear/ResumePriorityBoosting
Refers to Issue #550


MicroConsult:
Sorry for the mpu conflict. Please pull the core_cm33.h only.

JonatanAntoni:
@MicroConsult, 

Thanks for the patch. 

You could try to rework (using rebase, or cherry-picking) your patch and the two other unneeded commits if possible. The pull-request cannot be applied directly (by simply accepting it) while conflicting.

Cheers,
Jonatan

CMSIS/Core/Include/core_cm33.h
```diff
@@ -2083,7 +2089,92 @@ typedef struct
...
+  \return  0  Interrupt Priority Boosting was disabled
+  \return  1  Interrupt Priority Boosting was enabled
+ */
+__STATIC_INLINE uint32_t __NVIC_SetPriorityBoosting(void)
+{
+  uint32_t reg_value;
+  uint32_t prev_value;
+    
+  prev_value =  __NVIC_GetPriorityBoosting();                   /* remember old PRIS bit setting to be returned */
+
```
JonatanAntoni: I think we have an atomicity issue, here. Using this function from an RTOS might cause race conditions, i.e. the set/clear-restore cycle might not work as expected under all circumstances.

MicroConsult: While follwing the proposal from #550, I did not think about that topic, but you are pretty right. Those race conditions can easily be avoided when each thread stores its own history locally. Consequently we should switch back to the original proposal, which did not return the previous state and did not had a restore function.

JonatanAntoni: Yes, skipping the feature would of course avoid creating that kind of trap.
Alternatively we should think about making the get/set access atomic, i.e. using LDREX and STREX exclusive instructions.

MicroConsult: Using LDREX and STREX were not the problem, I just think about what should happen, if the STREX detects a conflict. Thus we did need to add an additional return value, since the conflict needs to be solved on the caller side of the new functions. e.g. to simply wait and retry in a loop. The entire race problem affects designs using TZ and an RTOS on the secure side. Its hard to predict but presumably most designs will use an RTOS on non-secure and bare metal on secure side? Is it then valid to complicate the interface? On the other hand one could think about two versions of the functions, simple without the restore for bare metal and EX as more complex and race-condition free with additional parameter to return potential conflicts?

JonatanAntoni: I expect TF-M (i.e. secure side "RTOS") to become "multi-threaded" one day as well. Not sure how this will look like. Working with exclusive load/store typically means repeat the sequence if it (the STREX) fails. That should resolve the race condition.

MicroConsult: Repetition in a loop does not solve the problem. You still might get the wrong old value delivered, which might have been changed by another thread in between. By applying the EX primitives you just get to know that a change happened in between, but no chnace to restore the old value. Thus the old value needs to be stored locally in eath thread. Thus EX does not help at all? I painted a picture that I wanted to attach as pdf, but github cmplained about it. I provided it for download via this link: [http://download.microconsult.net/rm/armv8m/RaceCondition.pdf](url)

JonatanAntoni: @MicroConsult: You're right. Using this from thread context seems to create wired situations we can hardly do correctly. If two threads change and restore the value of this bit while thread switching between both is allowed doesn't look deterministic at all. But then this set/clear/restore looks a bit useless to me. All one needs are simple get/set functions. One can "easily" implement set/clear/resume on top of that while taking care of atomicity.

MicroConsult: yes, we are now completly in line :-)
That is what I proposed on the start of this conversation. Leave away the restore function and leave away the return of the old value. That is further more what my first proposal in #550 did look like. So lets reverse to it. My proposal is to cancel/close this pull request and to setup a new one which contains the original #550 proposal and also does not contain the mpu staff not belonging to this pull request. Who can close/cancell this request? Is it me or you @JonatanAntoni  ?

JonatanAntoni: I guess you should be able to either update the existing pull request, or close it and create a new one. Updating might need to rewrite your branches history and force the push.

JonatanAntoni:
@MicroConsult, you still have merges and other stuff in between. Perhaps its easier to create a new PR and just cherry pick the relevant commit.

PR #564, opened 2019-04-03
Bugfix for NN_ROUND

freddan80:
For out_shift=0 the macro underflow


felix-johnny:
Other non related patches seem to have creeped in along with the actual fix.

PR #565, opened 2019-04-03
Fixing shift on zero issue

freddan80:
Input of zero gives bad output due to overflow.


JonatanAntoni:
Thanks for pointing this out. It looks reasonable to me.

PR #571, opened 2019-04-15
RTX5: Have osThreadJoin mark thread non-joinable

kjbracey-arm:
By inspection, this will have the following effects:

* A second call to osThreadJoin while the first is still waiting will now
  return with osErrorResource, rather than starting a join on this thread
  and leaving the first thread to never wake - we trap and avoid this
  application-caused deadlock.
* A call to osThreadDetach while an osThreadJoin is still waiting will
  now return with osErrorResource; this previously would have cleared
  the joinable flag and returned osOK.
* svcRtxThreadExit and svcRtxThreadTerminate implementations can be
  slightly simplified, as they only need to check the Joinable flag to
  decide whether to immediately free.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_thread.c
+++ b/CMSIS/RTOS2/RTX/Source/rtx_thread.c
@@ -1184,6 +1184,7 @@ static osStatus_t svcRtxThreadJoin (osThreadId_t thread_id) {
     // Suspend current Thread
     if (osRtxThreadWaitEnter(osRtxThreadWaitingJoin, osWaitForever)) {
       thread->thread_join = osRtxThreadGetRunning();
+      thread->attr &= ~osThreadJoinable;
       EvrRtxThreadJoinPending(thread);
     } else {
       EvrRtxThreadError(thread, (int32_t)osErrorResource);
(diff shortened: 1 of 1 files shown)


kjbracey-arm:
Spotted while fiddling with Mbed OS's `Thread` class, and considering use of `osThreadJoin`.

CMSIS-RTOS API doesn't say anything about what should happen if two threads try joining another thread, but clearly the design requires that only 1 do so. At the moment if multiple threads join another one, all will block forever except the last.

RobertRostohar:
You are correct, CMSIS-RTOS API does not say what should happen if multiple threads simultaneously try to join the same thread.

Join is not intended to be used in that way and the results are undefined. This should be mentioned in CMSIS-RTOS API documentation.

POSIX (and linux) also state that results are undefined is such case.

So from that perspective this is a user programming.

Anyway we will review the proposed pull request since it improves the robustness.

Thanks for raising this.

RobertRostohar:
PR has been reviewed, tested and looks good.

CMSIS-RTPS API documentation has also been updated with a note about the Join restriction.

Thanks for your help.

PR #572, opened 2019-04-17
New NEON float32 algorithms for Cortex A5 with armDS project

clamar01:
- Added in `cmsis_armclang.h` and `cmsis_gcc.h` from the `Core_A` directory the missing symbols that were present in the `Core` directory header files.
- Modified `arm_math.h` to support NEON
- Added the armDS project (`CMSIS/DSP/DSP_Lib_TestSuite/DspLibTest_FVP_A5`) to test the NEON float32 functions for A5
-- In `jtest_cycle.h` the function to test is called without counting the cycles for A5 (`ifdef ARMv7A`)
-- In `jtest_systick.h` no header file is needed for A5 since we are not counting the cycles (`ifdef ARMv7A`)
-- In `jtest_fw.h` added `printf` in dump functions since the project for A5 works on armDS (`ifdef ARMv7A`)


CMSIS/DSP/Include/arm_math.h
```diff
@@ -306,6 +306,9 @@
   #define ARM_MATH_CM0_FAMILY            1
 #elif defined (__ARM_ARCH_8M_MAIN__)
 //#define ARM_MATH_CM0_FAMILY            0
+#elif defined (ARM_MATH_NEON)
```
JonatanAntoni: This preprocessor condition does not fit the others. I don't expect ARM_MATH_NEON to be a predefined setting set by the compiler. Can we have this in a generic way, i.e. automatically deduced?

christophe0606: We will switch to a dsp_config.h in a future commit.

ua1arn: Please add this pair of strings as placeholder for use
DSP librady with Cortex-A9:
_#elif defined (__ARM_ARCH_7A__)
//#define ARM_MATH_CM0_FAMILY            0_

For NEON detecting wee can use __ARM_NEON symbol

PR #577, opened 2019-04-23
NVIC_SetVector: Add DSB to newer cores

### LINKED ISSUE #576: Should NVIC_SetVector have a DSB?
I saw some code for Cortex-M7 where someone had done:

        NVIC_SetVector(irq_n, vector);
    #if defined (__ICACHE_PRESENT) && (__ICACHE_PRESENT == 1U)
        SCB_InvalidateICache();
    #endif
        NVIC_EnableIRQ(irq_n);

I queried the invalidate, and they said they had seen a crash without it.

Now, I believe their actual issue is that you do architecturally need a `DSB` between setting a vector and enabling it, although ARM App Note 321 says M0, M0+, M1, M3 and M4 can manage without.

I think we should do one of the following, but I can't decide which:

* Add `DSB` to all `NVIC_SetVector`
* Add `DSB` to all `NVIC_SetVector` except M0,M0+,M1,M3,M4 (because we believe App Note 321)
* Add `DSB` to all `NVIC_EnableIRQ`
* Add `DSB` to all `NVIC_EnableIRQ` except M0,M0+,M1,M3,M4
* Leave it and document it with a note in each call.

(I do also have #493 open suggesting adding a compiler barrier to `NVIC_EnableIRQ` - putting a `DSB` there would supersede that. I think I favour putting the `DSB` in `NVIC_SetVector` or nowhere anyway).

kjbracey-arm:
Architecturally, a barrier is required after a vector table change to ensure that the newly-written entry is used by an immediately-occurring exception.

Without a barrier, the following sequences could fail:

    NVIC_SetVector(irq, irq_handler);
    NVIC_EnableIRQ(irq);

or

    NVIC_SetVector(SVCall_IRQn, svc_handler);
    __SVC(0);

A DSB at the end of `NVIC_SetVector` covers both cases.

A failure due to lack of barrier in the first case has been observed on Cortex-M7.

For earlier Cortex-M cores, ARM Application Note 321 ("ARM Cortex-M Programming Guide to Memory Barrier Instructions") states:

    ARM recommends that the architectural requirements are adopted.

    ... architecturally a memory barrier instruction is
    required after the vector table entry is updated, and if the
    exception is to be activated immediately.

    ...

    In the Cortex-M processors, omission of the DSB or DMB instructions
    does not cause any issue in vector entry modifications.

We take this last statement to apply to the M0, M0+, M1, M3 and M4 that the document covers, assuming further that it applies to all revisions beyond those listed, and also to the SC000 and SC300 (as variants of the M0 and M3).

ARM has not published similar implementation specifications for later cores, so a DSB is added to M7, M23, M33, M35P, and the generic ARMv8-M variants.

Fixes #576.

### DIFF
--- a/CMSIS/Core/Include/core_armv81mml.h
+++ b/CMSIS/Core/Include/core_armv81mml.h
@@ -2480,6 +2480,7 @@ __STATIC_INLINE void __NVIC_SetVector(IRQn_Type IRQn, uint32_t vector)
 {
   uint32_t *vectors = (uint32_t *)SCB->VTOR;
   vectors[(int32_t)IRQn + NVIC_USER_IRQ_OFFSET] = vector;
+  __DSB();
 }
 
 
--- a/CMSIS/Core/Include/core_armv8mbl.h
+++ b/CMSIS/Core/Include/core_armv8mbl.h
@@ -1552,6 +1552,7 @@ __STATIC_INLINE void __NVIC_SetVector(IRQn_Type IRQn, uint32_t vector)
   uint32_t *vectors = (uint32_t *)0x0U;
 #endif
   vectors[(int32_t)IRQn + NVIC_USER_IRQ_OFFSET] = vector;
+  __DSB();
 }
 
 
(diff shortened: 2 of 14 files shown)


kjbracey-arm:
It's come to my attention that if you don't have a `DSB` you still in principle need a compiler barrier - the following sequence could go wrong:

    NVIC_SetVector(irq, irq_handler);
    dev->IRQEN = SOMEIRQ_EN;

`dev` is presumably volatile, but that's not enough to make sure the vector write happens first. Either the vector table needs to be volatile, or there needs to be a barrier. Barrier seems to be the correct solution.

If #510 was in, it could use the __COMPILER_BARRIER from there.

But from experimentation while doing #493, #510, I've yet to see the lack of barrier cause an issue - general aliasing caution seems to be enough to keep accesses apart there, so maybe too here.

PR #582, opened 2019-04-29
Add Cortex-A9 and LTO optimization support

XGudron:
Allows you to use the library with a family of processors ARMv7-A (Cortex-A9).
Also, support for LTO optimization has been added, which can significantly reduce the amount of code when using the GCC compiler.


christophe0606:
Hi,

Thanks for the pull request and the work. Sorry for the late answer but as you may have remarked we are a bit bandwidth constrained on our side.

As a consequence, we prioritize bug report and new signal processing features for CMSIS-DSP. We can't spend time on build issues or build improvements except if they help on the long term.

That being said, there are a few issues :
1 - As far as I understand you worked with MDK ? But Cortex-A support should not be added to MDK project but only ArmDS project. We have recently introduced support for Cortex-A (and Neon code). So you can look at the ArmDS project we have added ;

2 - You have added Cortex-A9 support but I don't see work done to run the DSP Tests on Cortex-A (which should not be possible with MDK). We cannot add support for a new core without having possibility to retest with the standard framework. In the ArmDS project, you should be able to run the tests on Cortex-A. (FFT not yet since we are looking at best way to have Neon support here);

3 - You have introduced LTO optimization but also other optimizations flags. We cannot add optimizations flags without benchmarking again all functions because CMSIS-DSP is sometimes used in benchmarks.

4 - We generally regenerate the binaries only for official release. For development we don't want to merge binaries for the libraries ;

5 - I totally understand the need of modifying and tuning the project released with CMSIS-DSP since we cannot answer all the needs of the community. That's why we are working on a cmake build for CMSIS-DSP. Currently it is only supporting Arm AC6 compiler but with additional work it should not be too difficult to extend to gcc. I believe it will make it easier to customize the CMSIS-DSP build : different compilation flags, toolchains or customization of what is really included in the library. I would suggest to go in that direction if you need to change the build for your needs.

Thanks for your proposal and work. But, for all above reasons, we won't merge this pull request. I hope it makes sense to you.

Issue #541, opened 2019-02-26
Deploying a 1D Convolutional Network from Keras


atrevely:
I've been reading through this repository (and other examples) thoroughly but I'm still unclear on a few particulars related to deploying a neural network with 1D convolutional layers trained in Keras (Tensorflow backend).

1. I assume that 1D convolutional layers can be implemented using the arm_convolve_HWC_q...() functions? For N channels with length M, I assume it'd be identical to a 2D convolution with shape (M, N, 1)

2. I'm still very unclear about how to translate the weight matrix from Keras to one of these layers. If I have data with 2 channels and, say, 4 filters each with kernel size 3, I end up with a weight matrix that's (3, 2, 4). How should this get flattened? In Keras the weight matrix looks as follows:

[[[a1, a2, a3, a4],
 [b1, b2, b3, b4]],

[[c1, c2 , c3,  c4],
 [d1, d2, d3, d4]],

[[e1, e2, e3, e4],
 [f1, f2, f3, f4]]]

Does this turn into:

#define CONV1_WT {a1, a2, a3, a4, b1, b2, b3, b4, c1, ...}?

3. Now there's the quantization part. I assume the whole layer (after flattening) gets quantized together within the same range, but determining the bias_shift and out_shift is killing me, especially for convolutional layers. Is there a straightforward way of deriving the shifts during quantization?

4. Lastly, in Keras I'd use a flatten layer in order to move from the convolution to a dense layer. Does this need to be done by hand?

I realize there's a lot here, but any help with any of these issues would be very helpful for getting my working network over to an M4.

majianjia:
From my understanding, 

---

**(1)** You can use those "none-square" 2D for 1D data. simply set H shape to 1 should work. you should use (1, M, N) instead of your what you have said (M, N, 1). 

---

**(2)** Directly flattens the weights will not give you the format for CMSIS-NN, because they are stored in different orders. The weights in keras is (H, W, C, N) in 2D. N is the num of data. So you need this to convert to (N, H, W, C). When you use 1D, weights in keras is (W, C, N), then you need to reorder to (N, W, C) for CMSIS-NN
`cmsisnn_weights = np.transpose(keras_weights, (3, 0, 1, 2))` for 2D
`cmsisnn_weights = np.transpose(keras_weights, (2, 0, 1))`  for 1D

after that, you can store it to file directly, numpy will help you to flatten. 
[...]

Issue #455, opened 2018-10-19
feature request: nonsquare pooling functions


iabdalkader:
Hi, I have a person detection network trained on 64x128 dataset, I used the nonsquare convolution functions but the pooling functions crash because they don't support nonsquare input. I would like to request nonsquare max/ave pooling functions to work with the nonsquare convolution functions. I implemented a slow (non-dsp version) nonsquare pooling, but a dsp version will be much appreciated. Thanks!

saseptim:
As far as I can tell, the pooling function signature should be adjusted to include both x and y dimensions for pool, stride, pad, etc, and the code slightly adjusted such that dim_im_in should be changed to dim_im_in_x or dim_im_in_y as needed for example.

majianjia:
Same here. I am looking for 1D pooling, but finally, I have to make my own changes.  
All I have done is copy and paste the original one and delete all the loop with the y-axis. 
I believe you could modify it easily since the x and y are performed separately if you see the comments in the source

kwagyeman:
Would be nice in ARM fixed this.

PR #586, opened 2019-05-09
NN non-square and 1-D pooling

quinnabrvau:
Proposed four reference functions for non-square and 1D pooling (Max and Ave). It doesn't make a bunch of sense to have non-square CNNS but not non-square pooling as referenced in issue 455 ( #455 ) Also added the 1D convolution because it can be done slightly faster by separating it into its own function.
I also added 1D wrappers around the non-square CNN functions to allow simplified code for 1D CNNs to work with the 1D pooling. This would make issue 541 simpler. ( #541 ) I didn't seperate the CNN's into their own functions because there was not an obvious potential to save OPs 

*Note* I did add test to the test code but I don't have Keil so I haven't tested my implementations so there are probably bugs in the shapes of the Non-Square functions


quinnabrvau:
Sorry I broke this on accident, I'm going to reopen it

PR #589, opened 2019-05-15
Added register save to "__FPU_Enable()"

d-kato:
Fixed the values of registers R1 to R3 may be destroyed when inline expansion.
The problem was confirmed when the optimization was set to "-Os" in ARMC6.
The ARMC5 __ASM function automatically adds register saving even when inline expansion, but explicitly adds push and pop to match the processing with other compilers.

### DIFF
--- a/CMSIS/Core_A/Include/cmsis_armcc.h
+++ b/CMSIS/Core_A/Include/cmsis_armcc.h
@@ -481,6 +481,8 @@ __STATIC_INLINE __ASM void __FPU_Enable(void)
 {
         ARM
 
+        PUSH    {R1-R3}
+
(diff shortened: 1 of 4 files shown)


JonatanAntoni:
Hi  @d-kato,

thanks for pointing out this issue.

After discussing with some compiler experts I'd rather fix the inline assembly usage by adding proper clobber marker. Furthermore I think we can reduce the overhead when only using two instead of three registers.

Cheers,
Jonatan

JonatanAntoni:
@d-kato, may I ask you to double check if the provided fix in 3f3b67f fixes the issue for you?

d-kato:
Thank you very much. I confirmed that the issue was fixed.
So my pull request is closed.

PR #590, opened 2019-05-15
Newlib allocates stack size biger than 0x400 by default

alfedotov:
Current Newlib crt0.S implementation assumes stack size more than 0x400
See https://cygwin.com/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=libgloss/arm/crt0.S;h=1deb73aa5f66982848f5a3808d5a94dbe01a81e0;hb=HEAD#l201

### DIFF
--- a/Device/ARM/ARMCM0/Source/GCC/startup_ARMCM0.c
+++ b/Device/ARM/ARMCM0/Source/GCC/startup_ARMCM0.c
@@ -70,7 +70,11 @@ void Reset_Handler  (void) __attribute__ ((noreturn));
 //<h> Stack Configuration
 //  <o> Stack Size (in Bytes) <0x0-0xFFFFFFFF:8>
 //</h>
+#if defined (__NEWLIB__)
+#define  __STACK_SIZE  0x0001D000
+#else
 #define  __STACK_SIZE  0x00000400
+#endif
 static uint8_t stack[__STACK_SIZE] __attribute__ ((aligned(8), used, section(".stack")));
 
 //<h> Heap Configuration
--- a/Device/ARM/ARMCM0plus/Source/GCC/startup_ARMCM0plus.c
+++ b/Device/ARM/ARMCM0plus/Source/GCC/startup_ARMCM0plus.c
@@ -70,7 +70,11 @@ void Reset_Handler  (void) __attribute__ ((noreturn));
 //<h> Stack Configuration
 //  <o> Stack Size (in Bytes) <0x0-0xFFFFFFFF:8>
 //</h>
+#if defined (__NEWLIB__)
+#define  __STACK_SIZE  0x0001D000
+#else
 #define  __STACK_SIZE  0x00000400
+#endif
 static uint8_t stack[__STACK_SIZE] __attribute__ ((aligned(8), used, section(".stack")));
 
 //<h> Heap Configuration
--- a/Device/ARM/ARMCM1/Source/GCC/startup_ARMCM1.c
+++ b/Device/ARM/ARMCM1/Source/GCC/startup_ARMCM1.c
@@ -70,7 +70,11 @@ void Reset_Handler  (void) __attribute__ ((noreturn));
 //<h> Stack Configuration
 //  <o> Stack Size (in Bytes) <0x0-0xFFFFFFFF:8>
 //</h>
+#if defined (__NEWLIB__)
+#define  __STACK_SIZE  0x0001D000
+#else
 #define  __STACK_SIZE  0x00000400
+#endif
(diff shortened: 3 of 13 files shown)


JonatanAntoni:
Hi @alfedotov,

Thanks for pointing this out.
Can you explain what goes wrong with having only 1kb of reserved for stack by default?
Please bare in mind the files are only templates intended to be changed to the user needs anyway.

Having multiple variants for that define might cause other issues. Perhaps its a better approach to just raise an error if the stack configuration doesn't meet the requirements?

Cheers,
Jonatan

alfedotov:
>Can you explain what goes wrong with having only 1kb of reserved for stack by default?
Because Newlib takes StackTop value and allocates piece-by-piece stacks for different modes (FIQ/IRQ/USR etc ) or just allocates 0x10000 for User mode only.

> Perhaps its a better approach to just raise an error if the stack configuration doesn't meet the requirements?
You suggest something like this ?
```
#define STACK_SIZE 0x400
#ifdef __NEWLIB__ && STACK_SIZE < 0x1D000
#errior "Newlib requires more Stack size"
#endif
```

RobertRostohar:
> Because Newlib takes StackTop value and allocates piece-by-piece stacks for different modes (FIQ/IRQ/USR etc ) or just allocates 0x10000 for User mode only.

Cortex-M devices do not have the mentioned modes (only thread and handler mode).

JonatanAntoni:
@alfedotov,

That doesn't sound reasonable tome. You updated the startup code for Cortex-M based devices. These class of devices does not have the different modes you refer to. FIQ/IRQ/USR etc. are modes used on Cortex-A. Furthermore only some Cortex-M based devices provide at least 128KB of RAM. It doesn't look viable to me to have such high stack requirements on those small devices.

For Cortex-A devices in turn there is a memory config file which specifies the stacks specifically for all different modes. The stack pointers for all modes are initialized by the reset handler before jumping to library startup.

Can you please be more specific which exact problem you try to solve?

Cheers,
Jonatan

alfedotov:
> That doesn't sound reasonable tome. You updated the startup code for Cortex-M based devices

Indeed, I have forgot about this. I ran into this problem with Cortex-R52 core though.

>Cortex-M devices do not have the mentioned modes (only thread and handler mode).

That's true. But at the same newlib sets up Stack Limit register (for APCS) considering stack size is 64K https://cygwin.com/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=libgloss/arm/crt0.S;h=1deb73aa5f66982848f5a3808d5a94dbe01a81e0;hb=HEAD#l237

JonatanAntoni:
Hi @alfedotov,

Ah, okay. Now it's getting clear. Cortex-R series is not yet officially supported.

The portion of code you refer to looks a bit odd to me. I think stack sizes should be user configurable and not hard coded in any part of the library. I am not sure if this version of newlib is appropriate to be used in a microconroller environment.

Of course its up to the user to select the software components and libraries that fits the requirements best. I think we should not bind CMSIS tightly to any specific choice. The startup code provided is intended to give a viable starting point. The user should provide suitable configuration values on a per project basis.

Does this make sense to you?

Cheers,
Jonatan

alfedotov:
> Ah, okay. Now it's getting clear. Cortex-R series is not yet officially supported.

Any plans to support R profile in future ? We have some sources for R52 based on CMSIS  and can push here :)

> The portion of code you refer to looks a bit odd to me. I think stack sizes should be user configurable and not hard coded in any part of the library. I am not sure if this version of newlib is appropriate to be used in a microconroller environment.

Exactly. That's the major reason why I started to refactor crt0 code. Particularly for stack initialization change I'm trying to upstream is https://github.com/alfedotov/newlib/commit/75d143fcc32ed9779f15644e26cd285c033ec21d

JonatanAntoni:
Hi @alfedotov,

Yes, I think it would be beneficial to have Cortex-R support as well.

Personally I've not taken a closer look into it, yet. Hence I am not finally sure whether we can simply reuse what we have for Cortex-M. Or if we should add a separate Core_R folder. I assume the latter one might be needed to clearly distinguish the features.

Basically what would be needed to support Cortex-R class devices would be:
1) Add CMSIS/Core_R/*.h in a similar way as its in Core and Core_A.
2) Add some device startup templates (startup and system), i.e. Device/ARM/ARMCR52

Would you be able to provide some initial contribution for this?

Thanks,
Jonatan

alfedotov:
> Yes, I think it would be beneficial to have Cortex-R support as well.

Great :)

> Would you be able to provide some initial contribution for this?

Yes, it's one of my goals.

PR #596, opened 2019-05-21
Cortex r to develop

alfedotov:
This is minimal part of source code that is required to bring up Cortex-R52 and reach the main() routine.
For the moment there is no C variant of TCM and MPU initialization.
I've also added stubs for doxygen. In case if this wrong - please drop corresponding commit.


alfedotov:
Rebased to latest changes in develop branch (linker scripts, startup code etc)

CMSIS/Core_R/Include/cmsis_compiler.h
```diff
@@ -0,0 +1,213 @@
+/**************************************************************************//**
+ * @file     cmsis_compiler.h
+ * @brief    CMSIS compiler specific macros, functions, instructions
+ * @version  V1.0.2
+ * @date     10. January 2018
```
JonatanAntoni: We should set proper version and date information. I guess the file is simply a copy. But nevertheless the Cortex-R variant should be seen as a new file, i.e. starting at version 1.0.0

CMSIS/Core_R/Include/cmsis_cp15.h
```diff
@@ -0,0 +1,178 @@
+/**************************************************************************//**
+ * @file     cmsis_cp15.h
+ * @brief    CMSIS compiler specific macros, functions, instructions
+ * @version  
+ * @date
```
JonatanAntoni: Same here.

CMSIS/Core_R/Include/core_cr.h
```diff
@@ -0,0 +1,249 @@
...
+ extern "C" {
+#endif
+
+/*******************************************************************************
+ *                 CMSIS definitions
+ ******************************************************************************/
+
+/*  CMSIS CR definitions */
+#define __CR_CMSIS_VERSION_MAIN  (1U)                                      /*!< \brief [31:16] CMSIS-Core(R) main version   */
+#define __CR_CMSIS_VERSION_SUB   (1U)                                      /*!< \brief [15:0]  CMSIS-Core(R) sub version    */
```
JonatanAntoni: Version 1.0 should be the first one.

CMSIS/DoxyGen/Core_R/core_R.dxy
```diff
@@ -0,0 +1,2352 @@
...
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "CMSIS-Core (Cortex-R)"
+
+# The PROJECT_NUMBER tag can be used to enter a project or revision number. This
+# could be handy for archiving the generated documentation or if some version
+# control system is used.
+
+PROJECT_NUMBER         = "Version 1.1.4"
```
JonatanAntoni: Version again, should be 1.0.0 for the first release.

Device/ARM/ARMCR52/startup_ARMCR52.c
```diff
@@ -0,0 +1,159 @@
...
+  #include "ARMCR52.h"
+#else
+  #error device not specified!
+#endif
+
+/*----------------------------------------------------------------------------
+  Linker generated Symbols
+ *----------------------------------------------------------------------------*/
+extern uint32_t __HypStackTop;
+extern uint32_t __ecc_init_start__, __ecc_init_end__;
```
JonatanAntoni: These symbols look GCC-specific to me. We need to sort the compiler specifics out if we want to have the startup.c file agnostic right from the beginning. Another feasible solution would be adding FCC specific startup code, placing it to the GCC subfolder.

Device/ARM/ARMCR52/startup_ARMCR52.c
```diff
@@ -0,0 +1,159 @@
...
+void EL1_Abort_Handler    (void) __attribute__ ((weak, alias("EL1_Default_Handler")));
+void EL1_Reserved_Handler (void) __attribute__ ((weak, alias("EL1_Default_Handler")));
+void EL1_IRQ_Handler      (void) __attribute__ ((weak, alias("EL1_Default_Handler")));
+void EL1_FIQ_Handler      (void) __attribute__ ((weak, alias("EL1_Default_Handler")));
+
+/*----------------------------------------------------------------------------
+  Exception / Interrupt Vector table
+ *----------------------------------------------------------------------------*/
+void __VECTOR_TABLE_ATTRIBUTE __EL2_Vectors(void) {
+  __ASM volatile(
```
JonatanAntoni: This again is not really compiler agnostic as it uses GNU assembly syntax. Any chance to have this in C?

Device/ARM/ARMCR52/startup_ARMCR52.c
```diff
@@ -0,0 +1,159 @@
...
+
+  __ERET();
+
+  __builtin_unreachable(); /* No return here */
+}
+
+/*----------------------------------------------------------------------------
+  Reset Handler called on controller reset
+ *----------------------------------------------------------------------------*/
+void __attribute__((naked,__noreturn__)) EL2_Reset_Handler(void) {
```
JonatanAntoni: Naked functions implemented in C? At least in ArmClang this won't work, I suspect.

alfedotov: We cant rely on that stack won't be used here since SP isn't initialized yet (no hardware SP setup as in M profile). I've seen already that ``register`` keyword doesn't guarantee this for you.

Device/ARM/ARMCR52/startup_ARMCR52.c
alfedotov: I don't have other compilers right on my plate. Do you have an example ?

Device/ARM/ARMCR52/startup_ARMCR52.c
alfedotov: ECC should be done no matter which compiler is used. I see only way to declare new symbols like __ECC_INIT_START and __ECC_INIT_END and define them in compiler dependent header.

Device/ARM/ARMCR52/startup_ARMCR52.c
JonatanAntoni: Yes, indeed. I leave the choice up to you whether you want to investigate in a generalized startup code (like I did it for M-class recently) or stay with compiler specific files (like it is for A-class) for the time being. Compiler specific code should go into a sub directory.

alfedotov:
Rebased current PR.
1. Moved startup_ARMCR52.c to GCC subfolder.
2. Added abstraction for vectors (__VECTOR_TABLE_EL2, __VECTOR_TABLE_EL1) to cmsis_gcc.h

alfedotov:
1. Moved Device/ARM/ARMCR52/system_ARMCR52.c -> Device/ARM/ARMCR52/Source/system_ARMCR52.c
2. Aligned linker script with latest version in M-profile
3. Removed some stuff in headers that is not tested (copy-pasted from M-profile)

alfedotov:
1. Updated cmsis_cp15.h

alfedotov:
Hi
Any chances for review ? I have next changes but don't want to rebase again...

JonatanAntoni:
It's a fair starting point. We need to schedule the change after the upcoming 5.6 release.

alfedotov:
@JonatanAntoni when 5.6 is planned to be released ? Will you create a branch for it ?

JonatanAntoni:
We are currently aiming for next week. But we have ongoing work for other related packs, internally. Hence the actual release might slip further.

alfedotov:
Ping

PR #624, opened 2019-07-09
extern keyword not required

rajszym:
the extern keyword is not required for the external function declaration


JonatanAntoni:
Hi @rajszym,

You are right `extern` keyword is not required. But I think it states clearly that the function is required to be defined somewhere else. 

Do you have any concerns about using the keyword this way? Does it cause inconvenience or warnings?

Cheers,
Jonatan

rajszym:
Ok, I understand. Of course, it does not cause any inconvenience or warning.

rajszym:
To the trash

PR #630, opened 2019-07-18
CMSIS-NN: Add function for q7 to q15 conversion with offset

felix-johnny:
1. Added function arm_q7_to_q15_with_offset.c

2. Replaced __SIMD32 macros in arm_q7_to_q15_no_shift.c
   as an improvement.

### DIFF
--- a/CMSIS/NN/Source/NNSupportFunctions/arm_q7_to_q15_no_shift.c
+++ b/CMSIS/NN/Source/NNSupportFunctions/arm_q7_to_q15_no_shift.c
@@ -30,81 +30,71 @@
 
 #include "arm_nnsupportfunctions.h"
 
-/**    
- * @ingroup groupSupport    
+/**
+ * @ingroup groupSupport
  */
 
-/**    
- * @addtogroup nndata_convert    
- * @{    
+/**
+ * @addtogroup nndata_convert
+ * @{
  */
 
-/**    
- * @brief Converts the elements of the Q7 vector to Q15 vector without left-shift 
- * @param[in]       *pSrc points to the Q7 input vector    
- * @param[out]      *pDst points to the Q15 output vector   
- * @param[in]       blockSize length of the input vector    
- * @return none.    
- *    
- * \par Description:    
- *    
- * The equation used for the conversion process is:    
- *   
- * <pre>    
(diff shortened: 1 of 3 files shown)


CMSIS/NN/Source/NNSupportFunctions/arm_q7_to_q15_no_shift.c
```diff
@@ -30,81 +30,71 @@
...
  */
 
 void arm_q7_to_q15_no_shift(const q7_t * pSrc, q15_t * pDst, uint32_t blockSize)
 {
-    const q7_t *pIn = pSrc;     /* Src pointer */
-    uint32_t  blkCnt;           /* loop counter */
+    const q7_t *pIn = pSrc;
+    uint32_t  blkCnt;
 
 #ifndef ARM_MATH_CM0_FAMILY
```
christophe0606: We are no more using this flag in arm_math.h and CMSIS-DSP.

Instead I would use:

#if defined(ARM_MATH_LOOPUNROLL) && defined (ARM_MATH_DSP)

Because the loop is obviously unrolled and contained intrinsics

CMSIS/NN/Source/NNSupportFunctions/arm_q7_to_q15_with_offset.c
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+void arm_q7_to_q15_with_offset(const q7_t *src,
+                               q15_t *dst,
+                               uint32_t block_size,
+                               q7_t offset)
+{
+    uint32_t  block_cnt;
+
+#ifdef ARM_MATH_CM0_FAMILY
```
christophe0606: Same comment. ARM_MATH_CM0_FAMILLY no more used

Instead I would write

#if !defined(ARM_MATH_LOOPUNROLL) || !defined (ARM_MATH_DSP)

CMSIS/NN/Source/NNSupportFunctions/arm_q7_to_q15_no_shift.c
felix-johnny: ok..got it. I'll update the patch and correct this in other files in future patches..

PR #631, opened 2019-07-18
CMSIS-DSP: Prevent off-by-one access past end of y-data buffer

rmull:


christophe0606:
Thanks for reporting and correcting this bug.

PR #635, opened 2019-07-25
CMSIS-NN : Added function for s8 depthwise convolution

felix-johnny:
This is a reference version and optimized version will follow.

### DIFF
--- a/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8.c
+++ b/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8.c
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2010-2019 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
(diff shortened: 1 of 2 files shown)


CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8.c
```diff
@@ -0,0 +1,134 @@
...
+                                 const int32_t output_offset,
+                                 const int32_t input_offset,
+                                 const int32_t output_activation_min,
+                                 const int32_t output_activation_max,
+                                 const uint16_t dilation_x,
+                                 const uint16_t dilation_y,
+                                 q15_t *buffer_a)
+{
+    int i_out = 0;
+    (void)dilation_x;
```
christophe0606: So we don't support dilatation with this kernel ?

felix-johnny: Not as a first step as optimization with dilation support would be different(data packing for im2col) from that of the non-dilation version and it is not seen as a must have for embedded class networks. If we see a requirement for supporting it in future for embedded class networks, we can incrementally add support it.

PR #643, opened 2019-08-01
Updated prebuilt DSP libraries after commit on FFT f32

clamar01:
Updated libraries because of [this commit](https://github.com/ARM-software/CMSIS_5/commit/98b25b623d949b26495e24ecd9f357721ee56a96). Otherwise the real FFT f32 for a length of 128 cannot work properly.


ua1arn:
How about add pre-built libraries for Cortex-A9 variations (NEON/no NEON) ?

clamar01:
> How about add pre-built libraries for Cortex-A9 variations (NEON/no NEON) ?

This pull request is not the right place for this kind of questions.

JonatanAntoni:
Not updating binary files mid of a release cycle to limit database growth.
The pre-built libs might be updated again just before next public release. But we are discussing whether to drop pre-built libraries completely or at least remove them from source code repo.
Please raise an issue if you have any concerns.

PR #648, opened 2019-08-06
CMSIS_NN: Optimized s8 depthwise conv for channel multiplier = 1

felix-johnny:
1. Added new function for s8 depthwise conv supporting TensorFlow
   micro framework.
2. Fixed incorrect comment in
   arm_depthwise_separable_conv_HWC_q7_nonsquare.

### DIFF
--- a/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8_opt.c
+++ b/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8_opt.c
@@ -1,0 +1,316 @@
+/*
+ * Copyright (C) 2010-2019 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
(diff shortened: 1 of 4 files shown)


CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8_opt.c
```diff
@@ -0,0 +1,317 @@
...
+{
+
+
+    /* Check input constraints input_ch == output_ch */
+    if (input_ch != output_ch)
+    {
+        return ARM_MATH_SIZE_MISMATCH;
+    }
+
+// #if defined(ARM_MATH_LOOPUNROLL) && defined(ARM_MATH_DSP)
```
christophe0606: It should be uncommented and the #if 1 below removed.

felix-johnny: ah..my bad. I'll fix it

PR #651, opened 2019-08-13
Added __EARLY_INIT provision for all currently supported Cortex cores

alfedotov:
Early init is a must for implementation where ECC for RAM (and probably TCM) must be initialized before Stack being used.
This PR just pulls in a provision for that. Separate pull request will extend __EARLY_INIT by optional ECC implementation for Cortex-m7 etc

### DIFF
--- a/CMSIS/Core/Include/cmsis_armcc.h
+++ b/CMSIS/Core/Include/cmsis_armcc.h
@@ -1,8 +1,8 @@
 /**************************************************************************//**
  * @file     cmsis_armcc.h
  * @brief    CMSIS compiler ARMCC (Arm Compiler 5) header file
- * @version  V5.1.1
- * @date     30. July 2019
+ * @version  V5.1.2
+ * @date     08. August 2019
  ******************************************************************************/
 /*
  * Copyright (c) 2009-2019 Arm Limited. All rights reserved.
@@ -109,6 +109,10 @@
 #endif
 
 /* #########################  Startup and Lowlevel Init  ######################## */
+
+#ifndef __EARLY_INIT
+#define __EARLY_INIT()
+#endif
 
 #ifndef __PROGRAM_START
 #define __PROGRAM_START           __main
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -1,8 +1,8 @@
 /**************************************************************************//**
  * @file     cmsis_armclang.h
  * @brief    CMSIS compiler armclang (Arm Compiler 6) header file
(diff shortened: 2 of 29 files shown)


JonatanAntoni:
Hi @alfedotov,

thanks for your contribution.
Can you please describe a bit more detailed why the `SystemInit()` provision does not work for the situations you have in mind? Is it because `SystemInit` is a function that already uses a call stack?

Cheers,
Jonatan

alfedotov:
@JonatanAntoni Exactly. Memeory must be initialized prior to any usage including Stack.
SystemInit implemented as regular C routine that doesn't fit.

CMSIS/Core/Include/cmsis_armcc.h
```diff
@@ -110,6 +110,10 @@
 
 /* #########################  Startup and Lowlevel Init  ######################## */
 
+#ifndef __EARLY_INIT
+#define __EARLY_INIT()
```
JonatanAntoni: Why did you add the `__EARLY_INIT()` provision to the compiler headers? Just to have them empty by default? Or do you expect the implementation to be compiler specific?

alfedotov:
Yes, I'd say both things. And this PR just adds a provision. By the second PR for cortex M this macro will be filled

JonatanAntoni:
So, if these are just there to have them empty by default the compiler headers are the wrong location. The compiler headers are used for abstracting compiler specific stuff, not device or application specifics.

More over an ECC feature looks very device specific to me. I am not aware of any Cortex-M (or A) processor core implementing ECC features which need specific initialization. Hence filling those macros on core level would be the wrong place as well.

Instead I expect the initialization to be device (or even board) specific. In this case (device specific) the enhancement needs to go to the device specific startup code, not to the generic template. If its board specific (externally connected ECC memory) it should be left to the user to adopt the startup code template according to the specific needs.

alfedotov:
> So, if these are just there to have them empty by default the compiler headers are the wrong location. The compiler headers are used for abstracting compiler specific stuff, not device or application specifics.
I have followed the same approach as __PROGRAM_START. Yes, for the moment it is mostly empty.  But in future this may by others, not only for ECC. I can see only that demand on such pre-init exists.

You have a lot of macro like FPU_PRESENT. Why all of such macro can't be propagated on all cores and each core specific header will set exact value to describe what features are present.

> If its board specific (externally connected ECC memory) it should be left to the user to adopt the startup code template according to the specific needs.
It isn't board specific. For example cortex-m7 and cortex-r52 according to https://developer.arm.com/ip-products/processors

Anyway, I give up to upstream these changes and will keep them in my forked repo.

PR #652, opened 2019-08-14
CMSIS-NN: s8 1x1 convolution

felix-johnny:
A simplified int8 version of the generic convolution for the
1x1 convolution case.

Supported framework: TensorFlow Lite micro

### DIFF
--- a/CMSIS/NN/Include/arm_nnfunctions.h
+++ b/CMSIS/NN/Include/arm_nnfunctions.h
@@ -139,8 +139,8 @@ extern    "C"
    * @param[in,out]   output     pointer to output tensor. format: [H, W, out_ch]
    * @param[in]       output_shift    pointer to per output channel requantization shift parameter.
    * @param[in]       output_mult     pointer to per output channel requantization multiplier parameter.
-   * @param[in]       out_offset      output tensor offset.
-   * @param[in]       input_offset    input tensor offset.
+   * @param[in]       out_offset      output tensor offset. Range: int8
+   * @param[in]       input_offset    input tensor offset. Range: int8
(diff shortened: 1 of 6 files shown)


CMSIS/NN/Include/arm_nnfunctions.h
```diff
@@ -445,6 +445,65 @@ extern    "C"
...
+   * @return     The function returns either
+   *                  <code>ARM_MATH_SIZE_MISMATCH</code> if argument constraints fail. or,
+   *                  <code>ARM_MATH_SUCCESS</code> on successful completion.
+   *
+   * @details
+   *   - Supported framework : TensorFlow Lite Micro
+   *   - The following constrains on the arguments apply
+   *      -# output_ch and input_ch are multiples of 2
+   *      -# padding equals 0
+   *      -# Stride equals 1
```
freddan80: If pad and stride must be 0 and 1 respectively, what have the parameters in the interface?

felix-johnny: @freddan80  If we start with kernel_x and kernel_y parameters, it is a function intended for  1x1 convolution, and as the name of function and file has it, it is omitted from the argument list . i.e, quite obvious and is the base functionality of this function.

But, the assumptions/requirements for padding and stride values are at a level lower which I am not sure is always the case(0 and 1) when using a 1x1 convolution. So, for that minor reason and mainly to avoid a case where it is deployed and is accidentally/silently used for non compliant stride and pad values, it is included in the argument list.

felix-johnny: Rebased

freddan80: ok, fair enough

PR #659, opened 2019-08-19
CMSIS-DSP enable building python wrapper files for gnu linux

cdknorow:
This fixes the build issues, re enabling the pywrapper for the DSP library on linux

### DIFF
--- a/CMSIS/DSP/PythonWrapper/setup.py
+++ b/CMSIS/DSP/PythonWrapper/setup.py
@@ -14,7 +14,7 @@ if sys.platform == 'win32':
   # since the visual compiler and the win platform are
   # not supported by default in arm_math.h
 else:
-  cflags = ["-Wno-unused-variable","-Wno-implicit-function-declaration",config.cflags]
+  cflags = ["-Wno-unused-variable","-Wno-implicit-function-declaration",config.cflags,"-D__GNUC_PYTHON__"]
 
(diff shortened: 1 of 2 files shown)


cdknorow:
@christophe0606 this is based off your work, enabling the build for linux environment for python wrapper

CMSIS/DSP/PythonWrapper/setup.py
```diff
@@ -14,7 +14,7 @@
   # since the visual compiler and the win platform are
   # not supported by default in arm_math.h
 else:
-  cflags = ["-Wno-unused-variable","-Wno-implicit-function-declaration",config.cflags]
+  cflags = ["-Wno-unused-variable","-Wno-implicit-function-declaration",config.cflags,"-D__GNUC_PYTHON__"]
```
christophe0606: If setup.py is supporting windows and Linux then I think some logic should be added to detect it and set __GNUC_PYTHON__ only when building on Linux.

Otherwise it may break the build with MSVC.

christophe0606:
@cdknorow Thank a lot. I have added a little comment.

CMSIS/DSP/PythonWrapper/setup.py
cdknorow: @christophe0606  the cflags being set here are specifcally for linux. The first part of the if statement is doing the windows check.

christophe0606: @cdknorow Awesome. Thanks.

PR #663, opened 2019-08-28
CMSIS-NN : Added a table of s8 functions for quick reference.

felix-johnny:
List of functions that support s8 TFL specifiction is added along
with information like the additional memory that is required.

History is updated.

### DIFF
--- a/CMSIS/DoxyGen/NN/src/history.txt
+++ b/CMSIS/DoxyGen/NN/src/history.txt
@@ -7,6 +7,28 @@
     <th>Description</th>
   </tr>
   <tr>
+    <td>vTBD</td>
+    <td>
+      Added following functions for int8 operators with symmetric quantization.<br>
+      <ul>
+        <li>arm_convolve_s8</li>
+        <li>arm_depthwise_conv_s8</li>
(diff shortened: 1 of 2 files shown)


CMSIS/DoxyGen/NN/src/history.txt
```diff
@@ -6,6 +6,28 @@
     <th>Version</th>
     <th>Description</th>
   </tr>
+  <tr>
+    <td>vTBD</td>
+    <td>
+      Added following functions for int8 operators supporting TFL <br> symmetric quantization spec.<br>
```
freddan80: Where is the TFL spec? Couldn't this be considered a generic interface?

CMSIS/NN/README.md
```diff
@@ -0,0 +1,26 @@
+# CMSIS NN
+
+## About
+The table below gives  a quick overview of the API's in CMSIS-NN int8 library that support the TensorFlow Lite int8 specification.
```
freddan80: dito

felix-johnny: @freddan80 updated it now..

PR #670, opened 2019-09-02
cortex-m implementation of early_init

alfedotov:
This is the second part of EARLY_INIT support where optional ECC implementation for cortex-m7 is added.


alfedotov:
I'll keep these changes in my private repo.
Cheers

PR #673, opened 2019-09-03
CMSIS-NN : Added relu6 function

felix-johnny:
Pure C implementation applicable for all M cores.

Change-Id: I387eb79cca376f82ab82a212122dbf62895e5b5b

### DIFF
--- a/CMSIS/NN/Source/ActivationFunctions/arm_relu6_s8.c
+++ b/CMSIS/NN/Source/ActivationFunctions/arm_relu6_s8.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2010-2019 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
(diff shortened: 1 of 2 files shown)


CMSIS/NN/Source/ActivationFunctions/arm_relu6_s8.c
```diff
@@ -0,0 +1,65 @@
...
+   *
+   */
+
+void arm_relu6_s8(q7_t *data, uint16_t size)
+{
+    int32_t i;
+
+    for (i = 0; i < size; i++)
+    {
+        int32_t ip = data[i];
```
christophe0606: Why int32_t since in/out are q7 ?

felix-johnny: Doing operations in the native data type takes less instructions than doing it in short or char as it would otherwise take more instruction to do the conversion to int8 to int32 and back.

At least for the opeation  ip = MAX(ip, 0); , that save can be realized and from the cycle measurements that I did, the save was clear.

PR #686, opened 2019-09-23
Fix compilation define for RFFT F32 64

DrWelcome:
Define ARM_TABLE_REALCOEF_F32 instead of ARM_TABLE_REALCOEF_Q31 for RFFT F32 64


christophe0606:
Thank you for finding this issue.

Since I have moved the fft.cmake file into the DSP/Source folder, I won't merge directly this request.

But I have corrected the problem in my latest commit : https://github.com/ARM-software/CMSIS_5/commit/7814e32299d976eff4adc5181a104fa41e96c14a

PR #689, opened 2019-09-26
CMSIS-NN: Get optimized implementation temporary buffer sizes

freddan80:
Adding interface to get required buffer size for the operators that need it.


CMSIS/NN/Include/arm_nnfunctions.h
```diff
@@ -148,6 +148,7 @@ extern    "C"
    * @param[in]       buffer_a    pointer to buffer space used for input optimization(partial im2col) and is necessary
```
christophe0606: I wonder if it should be "param[out]" here since the buffer is modified ?

CMSIS/NN/Include/arm_nnfunctions.h
```diff
@@ -470,6 +483,7 @@ extern    "C"
    * @param[in]      buffer_a  pointer to buffer space used for input optimization(partial im2col) and is necessary
```
christophe0606: Same remark for param[in]. param[out] ?

I won't add the same comment for other similar lines...

CMSIS/NN/Include/arm_nnfunctions.h
freddan80: Good question, a matter of definition I guess

PR #701, opened 2019-10-11
CMSIS-NN : MVE support for elementwise add

felix-johnny:
M-Profile Vector Extension support is added along
with a vector function for requantization.

### DIFF
--- a/CMSIS/NN/Include/arm_nnsupportfunctions.h
+++ b/CMSIS/NN/Include/arm_nnsupportfunctions.h
@@ -426,6 +426,52 @@ __STATIC_FORCEINLINE q31_t arm_nn_requantize(const q31_t val, const q31_t multip
                                        RIGHT_SHIFT(shift));
 }
 
+#if defined(ARM_MATH_MVEI)
+/**
+ * @brief           Vector saturating doubling high multiply returning high half.
+ * @param[in]       m1        Multiplicand
+ * @param[in]       m2        Multiplier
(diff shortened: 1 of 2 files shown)


felix-johnny:
Needs a rebase..

CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -426,6 +426,52 @@ __STATIC_FORCEINLINE q31_t arm_nn_requantize(const q31_t val, const q31_t multip
...
 
+#if defined(ARM_MATH_MVEI)
+/**
+ * @brief           Vector saturating doubling high multiply returning high half.
+ * @param[in]       m1        Multiplicand
+ * @param[in]       m2        Multiplier
+ * @return          Result of multiplication.
+ *
+ */
+__STATIC_FORCEINLINE int32x4_t arm_mve_sat_doubling_high_mult(const int32x4_t m1, const q31_t m2)
```
gmiodice: Why q31_t? Is it fine to keep using q31_t rather than int32_t ?

felix-johnny: @gmiodice It is a bit of a mix with q31_t and int32_t with q31_t more prevalent in the function arguments as the legacy API's that we still use are mainly q31_t. The q31_t is retained here as the non-vector API arm_nn_sat_doubling_high_mult() has it as q31_t.

PR #707, opened 2019-10-14
CMSIS-NN: Implement concatenation layer - Int8\Uint8

gmiodice:
This patch adds the support for concatenation int8/uint8 in CMSIS-NN


CMSIS/DSP/Testing/preprocess.py
```diff
@@ -1,18 +1,18 @@
...
-parser.add_argument('-o', nargs='?',type = str, default="Output.pickle", help="output file for parsed description")
-
-args = parser.parse_args()
-
-if args.f is not None:
-    p = parse.Parser()
-    # Parse the test description file
-    root = p.parse(args.f)
-    with open(args.o,"wb") as output:
-         pickle.dump(root, output)
```
gmiodice: This file has been changed because there was an issue with the new line character

felix-johnny: @gmiodice Looks good and nice with the pseudo code. Got one naming consideration comment. Channels are usually named as input_ch or output_ch in most other operators.. Maybe you could change just that so it is uniform across files?

gmiodice: Since these functions are data layout independent, I add a comment in the doxygen documentation.

gmiodice: Since these functions are data layout independent, I added a comment in the doxygen documentation

PR #708, opened 2019-10-16
CMSIS-NN: Adding batch to conv kernel

freddan80:


CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_1x1_s8_fast.c
```diff
@@ -79,102 +80,107 @@ arm_status arm_convolve_1x1_s8_fast(const q7_t *input,
         return ARM_MATH_SIZE_MISMATCH;
     }
 
+    int i_batch;
+    for (i_batch = 0; i_batch < input_batches; i_batch++)
+    {
+        input += i_batch * (input_x * input_y * input_ch);
+        output += i_batch * (output_x * output_y * output_ch);
```
Tombana, line 87: I was wondering if these lines should actually be one of the following:
```cpp
input = input_start + i_batch * (input_x * input_y * input_ch);
```
or
```cpp
input += (input_x * input_y * input_ch);
```
(and similar for `output`)

felix-johnny: @Tombana .. You are right.. I think it was fixed quite immediately after this was merged.
 We don't loop over the batches for 1x1 conv anymore though.. 

Here is the latest version of the file https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_1x1_s8_fast.c

Tombana: Thank you for the quick response!

I actually meant to comment on `arm_convolve_s8.c` in my original comment but I mixed up the files. In `arm_convolve_s8.c` the problem still exists on the develop branch:
https://github.com/ARM-software/CMSIS_5/blob/a68c2a2febf50d45a718a6a7e1c0207e60253af7/CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_s8.c#L79-L80

felix-johnny: @Tombana A quick response followed by a late one! That is true. It is still an issue in arm_convolve_s8.c . Absolutely, you are welcome to make a PR on that. On my part I'll add a test case for handling batches > 2 which would have identified this.

Tombana: Thanks! See https://github.com/ARM-software/CMSIS_5/pull/909

PR #716, opened 2019-10-25
CMSIS-DSP: Added fixed point Neon BasicMath functions

clamar01:
All the fixed point function of the BasicMath group (q7, q15, q31) now have also the Neon version


christophe0606:
Reviews will continue internally.

PR #717, opened 2019-10-29
CMSIS-NN: MVE support for s8 Fully connected and MatMul

felix-johnny:
M-Porfile Vector Extension support is added.
Author: christophe0606


CMSIS/NN/Include/arm_nnfunctions.h
```diff
@@ -975,7 +975,8 @@ int32_t arm_depthwise_conv_s8_opt_get_buffer_size(const uint16_t input_ch,
    * @param[in]       output_activation_min        for clamping
    * @param[in]       output_activation_max        for clamping
    * @param[in]       vec_buffer                   pointer to buffer space used for optimization and is necessary
```
gmiodice: How big should vec_buffer be?

felix-johnny: @gmiodice It is mentioned couple of lines below as 'col_dim * sizeof(q15_t) bytes'

PR #722, opened 2019-11-06
CMSIS: Update c startup source code to support multi-file compilation mode in IAR

GorgonMeducer:
The previous version of Cortex-M c startup file cannot be used in IAR multi-file compilation mode which applies strict function and variable type check.

When users implement their own ISR functions, the function's default attribute conflict with the ISR functions defined in the c startup code, i.e. the existence of "alias" attribute.

By removing "alias" from the default ISR function attribute (while keeping the weak), the new c startup code can work with IAR multi-file compilation mode,  Arm Compiler 5, Arm Compiler 6 and GCC.

### DIFF
--- a/CMSIS/DSP/Platforms/MPS3/ARMv81MML/Include/ARMv81MML_DSP_DP_MVE_FP.h
+++ b/CMSIS/DSP/Platforms/MPS3/ARMv81MML/Include/ARMv81MML_DSP_DP_MVE_FP.h
@@ -1,0 +1,132 @@
+/**************************************************************************//**
+ * @file     ARMv81MML_DP.h
+ * @brief    CMSIS Core Peripheral Access Layer Header File for
+ *           Armv8.1-M Mainline Device Series (configured for Armv8.1-M Mainline with double precision FPU, with DSP extension, with TrustZone)
+ * @version  V1.0.0
+ * @date     25. February 2019
+ ******************************************************************************/
+/*
+ * Copyright (c) 2009-2019 Arm Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ARMv81MML_DSP_DP_H
+#define ARMv81MML_DSP_DP_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/* -------------------------  Interrupt Number Definition  ------------------------ */
+
+typedef enum IRQn
+{
+/* --------------------  Armv8.1-M Mainline Processor Exceptions Numbers  --------- */
+  NonMaskableInt_IRQn           = -14,      /*  2 Non Maskable Interrupt */
+  HardFault_IRQn                = -13,      /*  3 HardFault Interrupt */
+  MemoryManagement_IRQn         = -12,      /*  4 Memory Management Interrupt */
+  BusFault_IRQn                 = -11,      /*  5 Bus Fault Interrupt */
+  UsageFault_IRQn               = -10,      /*  6 Usage Fault Interrupt */
+  SecureFault_IRQn              =  -9,      /*  7 Secure Fault Interrupt */
+  SVCall_IRQn                   =  -5,      /* 11 SV Call Interrupt */
+  DebugMonitor_IRQn             =  -4,      /* 12 Debug Monitor Interrupt */
+  PendSV_IRQn                   =  -2,      /* 14 Pend SV Interrupt */
+  SysTick_IRQn                  =  -1,      /* 15 System Tick Interrupt */
+
+/* -------------------  Processor Interrupt Numbers  ------------------------------ */
+  Interrupt0_IRQn               =   0,
+  Interrupt1_IRQn               =   1,
+  Interrupt2_IRQn               =   2,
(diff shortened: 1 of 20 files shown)


TTornblom:
Do you have any simple test project that shows the problem?

I created a simple M3 project where I declare the following in main.c:
---
`
void HardFault_Handler()
{
}

main()
{
}
`
When I compile this with EWARM 8.40.2 in mfc mode I get a complaint about:
---
```
main.c startup_ARMCM3.c system_ARMCM3.c  
Error[Pe1061]: declaration of variable "__vector_table" is incompatible with a declaration in another translation unit C:\Users\thomasto\Projects\8.40\8.40.2\test3\test3\RTE\Device\ARMCM3\system_ARMCM3.c 40 
            "uint32_t __vector_table" (declared at line 40) (from translation unit "C:\Users\thomasto\Projects\8.40\8.40.2\test3\test3\RTE\Device\ARMCM3\ 
system_ARMCM3.c") 
            "pFunc const __vector_table[240] @ ".intvec"" (declared at line 81 of "C:\Users\thomasto\Projects\8.40\8.40.2\test3\test3\RTE\Device\ARMCM3\ 
startup_ARMCM3.c") (from translation unit "C:\Users\thomasto\Projects\8.40\8.40.2\test3\test3\RTE\Device\ARMCM3\startup_ARMCM3.c") 
          detected during compilation of secondary translation unit "C:\Users\thomasto\Projects\8.40\8.40.2\test3\test3\RTE\Device\ARMCM3\ 
system_ARMCM3.c" 
Error while running C/C++ Compiler 
 
Total number of errors: 1 
```

but nothing from the alias.

The issue above appears to be a mismatch where __VECTOR_TABLE is declared as:
`       const pFunc __VECTOR_TABLE[240] __VECTOR_TABLE_ATTRIBUTE = {
`
in startup_ARMCM3.c, but as:
`  extern uint32_t __VECTOR_TABLE;
`
in system_ARMCM3.c.

If I change this to:
---
```
typedef void( *pFunc )( void );
extern const pFunc __VECTOR_TABLE[];
```
then the project builds without errors and the map file shows that it is using the HardFault_Handler from main.c

JonatanAntoni:
Hi @TTornblom,

I guess this is yet another issue we need to address. But the original issue is about weak aliases. Hardfault Handler is the only one that has *no* weak alias in CM3. Please give it a try with any other fault or even with `Interrupt0_Handler` for instance. Those are all weak aliases to the `Default_Handler`.

Cheers,
Jonatan

TTornblom:
Yes, I'm aware that the original issue was about the weak aliases, but I am unable to reproduce this.
I've also tried with BusFault_Handler and Interrupt0_handler and cm0 and cm4 targets, the only problem I have with my trivial test is with the mismatched __VECTOR_TABLE between the startup_*.c and system_*.c files.
I would like an example that shows the problem this patch is solving.

GorgonMeducer:
[IAR_CMSIS.zip](https://github.com/ARM-software/CMSIS_5/files/3815176/IAR_CMSIS.zip)
Here is the reproducer.

TTornblom:
Thanks,
Changing:
```
void SysTick_Handler(void)
{

}
```
to
```
void SysTick_Handler()
{

}
apparently solves this, which is why I had no problems with my simple example.
```

GorgonMeducer:
Changing 

> Thanks,
> Changing:
> 
> ```
> void SysTick_Handler(void)
> {
> 
> }
> ```
> 
> to
> 
> ```
> void SysTick_Handler()
> {
> 
> }
> ```
> 
> apparently solves this, which is why I had no problems with my simple example.

- a. this method is not well documented
- b. ignoring "void" is not recommended in recent c standard.
- c. Most of customsers' code still use void SysTick_Handler(void). If we recommend them to do so, it only makes IAR more "different".
- d. If IAR thinks this is a "feature", then I would like to recommend removing "alias" to make the user implemented handler looking normal. 

Thank you for your time.

TTornblom:
The patch appears to solve this issue, so I'm fine with it.

I will go over the system_*.c files and fix the VECTOR_TABLE type mismatch.

GorgonMeducer:
> The patch appears to solve this issue, so I'm fine with it.
> 
> I will go over the system_*.c files and fix the VECTOR_TABLE type mismatch.

My patch also solves the problem, how about accepting my one.

TTornblom:
How does this patch solve the type mismatch for VECTOR_TABLE? In your example you have excluded system_ARMCM7.c from the build, so you are not running into this issue. If you don't do that then you will run into the issue.

This is a separate issue from the problem you have solved. I meant to say I was fine with your patch and I suggest merging it, and I will address the type mismatch for VECTOR_TABLE.

GorgonMeducer:
> How does this patch solve the type mismatch for VECTOR_TABLE? In your example you have excluded system_ARMCM7.c from the build, so you are not running into this issue. If you don't do that then you will run into the issue.
> 
> This is a separate issue from the problem you have solved. I meant to say I was fine with your patch and I suggest merging it, and I will address the type mismatch for VECTOR_TABLE.

I think I am talking different issue, I am talking the alias one. About the VECTOR_TABLE issue, I have no problem with that.

JonatanAntoni:
Hi @TTornblom,

can you please explain me a bit more in detail why removing the `void` makes a difference?
I don't get the problem or the solution, yet. 

Cheers,
Jonatan

TTornblom:
I have asked the compiler team to explain this.

JonatanAntoni:
Hi Thomas,

I did the same and got pointed to https://softwareengineering.stackexchange.com/questions/286490/what-is-the-difference-between-function-and-functionvoid.

This is actually a good write up what I already had in mind. There is actually a difference on the function declaration between `func()` and `func(void)` but **not** on the definition. 

The CMSIS startup code uses a distinct way of declaring all the interrupt handlers to have **no** arguments (by declaring it as `handler(void)`). Hence I expect the function definition should be `handler(void)` and not `handler()` even if both are the same.

Can you discuss with your compiler experts why this seems to be handled differently in IAR compiler, please?

Cheers,
Jonatan

TTornblom:
Turns out this is a parser bug in our compiler.

It will be fixed in a future version. Unfortunately there is no time to get it into the next release but hopefully in a patch release later.

GorgonMeducer:
Since it is an IAR parser issue which will be fixed in the future, I will close this PR.

PR #728, opened 2019-11-08
Device(M): Refactoring __VECTOR_TABLE to avoid MFC issues with IAR

TTornblom:
Moved the typedef of pFunc from startup_ARM*.c to system_ARM*.h.
Moved the extern declaration of __VECTOR_TABLE from system_ARM*.c to system_ARM*.h
Inserted parenthesis around the usage of __VECTOR_TABLE in system_ARM*.c

Have tested with both assembler and C startup files on IAR. Have not tested with gcc or mdk


Device/ARM/ARMCM0/Include/system_ARMCM0.h
```diff
@@ -30,6 +30,13 @@
 extern "C" {
 #endif
 
+/*----------------------------------------------------------------------------
+  Exception / Interrupt Handler Function Prototype
+ *----------------------------------------------------------------------------*/
+typedef void( *pFunc )( void );
+
+extern const pFunc __VECTOR_TABLE[48];
```
JonatanAntoni: Unfortunately, moving this declaration to the header file breaks backward compatibility to *all* existing projects using former startup code. We need to check if we can move this back to the implementation. Or we need to revert completely. Sorry. 
cc @GorgonMeducer

PR #731, opened 2019-11-10
CMSIS-NN: MVE support for s8 softmax

felix-johnny:


CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -349,8 +349,14 @@ void arm_nn_mult_q7(
 
 // Macros for shortening quantization functions' names and avoid long lines
 #define MUL_SAT(a, b)  arm_nn_sat_doubling_high_mult((a), (b))
+#define MUL_SAT_MVE(a, b) arm_mve_sat_doubling_high_mult_32x4((a), (b))
```
gmiodice: Just a question about naming: before we called that function with arm_nn prefix. Should we keep the prefix here as well and move mve at the end of the string?

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -48,12 +89,118 @@ void arm_softmax_s8(const int8_t *input,
...
+#define ACT_MIN (-128)
+#define ACT_MAX (127)
+
+    const int32_t mask = (1 << shift);
+
+    for (int i_num_rows = 0; i_num_rows < num_rows; ++i_num_rows)
+    {
+        int8_t max = ACT_MIN;
+
+        int32_t vec_count = row_size / 16;
```
gmiodice: Since the division could be expensive, what about:
        for (int i = 0; i <= (vec_size - 16); i+=16)
        {
            const int8x16_t ip = vldrbq_s8(&input[i]);
            max = vmaxvq_s8(max, ip);
        }

In this manner you can also avoid the multiplication by 16 for calculating the array index

CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -379,7 +385,7 @@ __STATIC_FORCEINLINE q31_t arm_nn_sat_doubling_high_mult(const q31_t m1, const q
     // as well.
     result = mult / (1UL << 31);
 
-    if ((m1 == m2) && (m1 == Q31_MIN))
+    if ((m1 == m2) && (m1 == (int32_t)Q31_MIN))
```
giorgio-arenarm: Fair point, though I think it'd be better to define it as a 32bit integer in arm_math.h directly (it makes more sense to cast it to long if needed)

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -28,10 +28,50 @@
  *
  * -------------------------------------------------------------------- */
 
-#include "arm_nnfunctions.h"
+#include "arm_nnsupportfunctions.h"
```
giorgio-arenarm: Why not arm_nnfunctions?

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -28,10 +28,50 @@
...
 
 #define ACCUM_BITS 12
 
+#ifdef ARM_MATH_MVEI
+int32x4_t arm_nn_exp_on_negative_values_32x4(int32x4_t val)
```
giorgio-arenarm: Shouldn't this go in arm_nnsupportfunctions.h together with all the other quant ops?

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -48,12 +89,118 @@ void arm_softmax_s8(const int8_t *input,
                     const int8_t diff_min,
                     int8_t *output)
 {
+#ifdef ARM_MATH_MVEI
+
+#define ACT_MIN (-128)
+#define ACT_MAX (127)
```
giorgio-arenarm: Q7_MIN/MAX? Softmax shouldn't have weird activations attached to it since it is the last layer used for classification

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -48,12 +89,118 @@ void arm_softmax_s8(const int8_t *input,
...
+        {
+            const int8x16_t ip = vldrbq_z_s8(&input[vec_count * 16], p);
+            max = vmaxvq_p_s8(max, ip, p);
+        }
+
+        vec_count = row_size / 4;
+        int32_t idx = 0;
+        int32_t sum = 0;
+
+        while (vec_count)
```
giorgio-arenarm: for loop?

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
```diff
@@ -48,12 +89,118 @@ void arm_softmax_s8(const int8_t *input,
...
+        }
+
+        const int32_t headroom = __CLZ(sum);
+        const int32_t bits_over_unit = ACCUM_BITS - headroom + 23;
+        const int32_t shifted_scale = ONE_OVER1((sum << headroom) - (1 << 31));
+
+        vec_count = row_size / 4;
+        idx = 0;
+
+        while (vec_count)
```
giorgio-arenarm: same

CMSIS/NN/Include/arm_nnsupportfunctions.h
felix-johnny: @gmiodice Good point.. Moving mve to the end - yes. A reason, I was eager to get rid of the _nn is, it has shown to cause some confusion with  ARM NN. How about arm_sat_doubling_high_mult() and arm_sat_doubling_high_mult_mve_32x4() ?

CMSIS/NN/Include/arm_nnsupportfunctions.h
felix-johnny: @giorgio-arenarm .. True.  @christophe0606 .. just a suggestion :-)

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
felix-johnny: arm_nnfunctions.h is the external API interface to CMSIS-NN which also includes arm_nnsupportfunctions.h. Including just arm_nnsupportfunctions.h is a way of saying that this is a standalone function that does not depend on any of the functions in the external API list.

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
felix-johnny: If it is used in more than one location/file, then yes. We can move it to the header file when that happens.

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
felix-johnny: @giorgio-arenarm Sure.. i'll fix that.

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
felix-johnny: @gmiodice For the division, I don't think so as it is  a divide by a power of 2 number which should be resolved as  a shift operation. Normally, I would have done it as row_size >> 4, but for clarity i have gone with /16 and trust the compiler to do the rest. With that in mind, do want to change to avoid the multiplication alone?

gmiodice: So it is ok to keep it as it is. ;)

CMSIS/NN/Source/SoftmaxFunctions/arm_softmax_s8.c
felix-johnny: just a developer's choice :-)

felix-johnny:
@gmiodice @giorgio-arenarm Are you ok with the updates to the comments?

gmiodice:
All good for me ;) Thanks for checking

PR #736, opened 2019-11-15
Fix elementwise multiplication when output shift is positive

giorgio-arenarm:


felix-johnny:
@giorgio-arenarm Thanks for the fix!

PR #737, opened 2019-11-17
Fixed DAP PIN_DELAY_SLOW() when highly optimized

### LINKED ISSUE #389: CMSIS DAP delay is optimized out by gcc with aggressive optimization options
In fuction PIN_DELAY_SLOW:
```
__STATIC_FORCEINLINE void PIN_DELAY_SLOW (uint32_t delay) {
  uint32_t count;

  count = delay ;
  while (--count);
}
```
the loop is getting optimized out with -O3 option.

**Proposed solution:** 
Make the `count` `volatile` or use something similar to this:
```
__STATIC_FORCEINLINE void PIN_DELAY_SLOW (uint32_t delay) {
  uint32_t count;

  count = delay ;
  while (--count)  __asm__("" : "=r" (delay));
}
```
which generates very efficient code:
288         while (--count)  __asm__("" : "=r" (delay));;
```
0800c5c8:   beq.w   0x800cac0 <SWD_TransferSlow+1296>
0800c5cc:   mov     r2, r3
0800c5ce:   subs    r2, #1
0800c5d0:   bne.n   0x800c5ce <SWD_TransferSlow+30>
```

flit:
The `PIN_DELAY_SLOW()` function in `DAP.h` was disappearing when built with gcc using any optimization setting but `-O0` or `-Og`. The cause is that the loop count variable needs to be volatile, otherwise an optimizing compiler is likely to elide the entire while loop.

Closes #389.

### DIFF
--- a/CMSIS/DAP/Firmware/Include/DAP.h
+++ b/CMSIS/DAP/Firmware/Include/DAP.h
@@ -280,9 +280,7 @@ extern void     DAP_Setup (void);
 #define DELAY_SLOW_CYCLES       3U      // Number of cycles for one iteration
 #endif
 __STATIC_FORCEINLINE void PIN_DELAY_SLOW (uint32_t delay) {
-  uint32_t count;
-
-  count = delay;
+  volatile uint32_t count = delay;
   while (--count);
 }
 


JonatanAntoni:
Relates to #389.

flit:
Closing due to 680e079.

PR #741, opened 2019-11-19
[CMSIS-DSP] Fix the wrong pivot row index in mat_inverse_f32().

JerryShih:
The mat_inverse_f32() function's result is not correct. I think it's just a typo for handling the zero pivot element.
I will get the same result with numpy mat_inverse() function with this change.
I don't update other code block since I don't have the corresponding environment to test.

Here is a simple test that will break the original implementation.
```
input 2x2 matrix:
0, 3
4, 5
The inv mat should be:
-0.41666667,  0.25,
0.33333333,  0
But we got:
-0.15, 0.25
0.2, 0
```

### DIFF
--- a/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f32.c
+++ b/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f32.c
@@ -1471,7 +1471,7 @@ arm_status arm_mat_inverse_f32(
         for (i = (l + 1U); i < numRows; i++)
         {
           /* Update the input and destination pointers */
-          pInT2 = pInT1 + (numCols * l);
+          pInT2 = pInT1 + (numCols * i);
           pOutT2 = pOutT1 + (numCols * k);
 
           /* Check if there is a non zero pivot element to


code2pi:
Really? I tried calling this function, it seems correct. Is this a special case?

JerryShih:
@code2pi 
Could you just try with the pure c implementation(no neon/mvef)?
And could you also try with the following input data?
mat 2x2:
0, 3
4, 5
The first pivot element is zero. That will enter the code block [link](https://github.com/ARM-software/CMSIS_5/blob/bb331775de083cda0971b47ede744f4129c62cd7/CMSIS/DSP/Source/MatrixFunctions/arm_mat_inverse_f32.c#L1467-L1508).

christophe0606:
Thanks for reporting this bug.
Since it is also impacting the unrolled, neon and helium versions, I'll commit a fix for all versions rather than merging this pull request.

JerryShih:
@christophe0606 
I'm not familiar with cmsis code-base. Where is the test case for this mat_inverse() function? How about the mat_inverse_f64()?

christophe0606:
@JerryShih  I have pushed a commit solving the issue and aligning the scalar and vectorial versions.
It is on our todo list to improve those functions for numerical stability.

The test cases are in DSP/DSP_Lib_TestSuite.

But this test suite is deprecated and we are working on better tests. But it is not ready yet.
I have added your matrix to the new tests so that this issue no more occur.

PR #749, opened 2019-11-25
Protect TRUE/FALSE defines for RTOS2

flit:
The file `rtx_core_cm.h` defines `TRUE` and `FALSE` macros. This change protects these definitions in case they are already defined. This can happen if the MCU's CMSIS register definition header defines such macros, which will included through the `#include CMSIS_device_header` line.

A much better fix would be to use the `true` and `false` definitions from `<stdbool.h>`, which is already included in this header. There is no need to define your custom bool types, especially since they are just redefined to the same as the ones in `<stdbool.h>`. If you'll accept such a proper fix, I'll make the changes.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_core_cm.h
+++ b/CMSIS/RTOS2/RTX/Source/rtx_core_cm.h
@@ -33,8 +33,14 @@
 
 #include <stdbool.h>
 typedef bool bool_t;
+
+#ifndef FALSE
 #define FALSE                   ((bool_t)0)
+#endif
+
+#ifndef TRUE
 #define TRUE                    ((bool_t)1)
+#endif
 
 #ifdef  RTE_CMSIS_RTOS2_RTX5_ARMV8M_NS
 #define DOMAIN_NS               1


RobertRostohar:
The reason for not using `true `and `false` is MISRA compliance. Using them directly results in MISRA warnings: "Use of modifier or type outside of a typedef [MISRA 2012 Directive 4.6, advisory]".

Therefore we have defined a custom type `bool_t` with `TRUE` and `FALSE` definitions.

RobertRostohar:
Adding guards to `TRUE` and `FALSE` definitions makes sense as you have pointed out. Thanks for raising this.

flit:
MISRA directive 4.6 seems to apply only to numeric types: The 4.6 headline is, "typedefs that indicate size and signedness should be used in place of the basic _numerical_ types" (see page 28 of MISRA 2012). It's also an advisory rather than required directive.

My experience with using linter tools for MISRA compliance is that they very often incorrectly report errors because they don't fully understand the context.

MISRA interpretation can be tricky… 😄 There are other MISRA directives that CMSIS-RTOS2 does not follow, such as 4.12 "Dynamic memory allocation shall not be used" (required).

flit:
Sorry I missed the same definitions in `rtx_core_ca.h`!

PR #750, opened 2019-11-25
DAP.h hygiene

flit:
Some tiny changes to `DAP.h`.

1. Move includes to the top.
2. Wrap API definitions with `extern "C"` for C++.


RobertRostohar:
There is no need to move the includes (would like to keep the order for historic reasons).

`extern "C"` has been added in a separate commit 56acd94. Thanks for raising this.

PR #754, opened 2019-11-27
Adding return value to pooling functions

freddan80:


felix-johnny:
@freddan80 You can merge it when the TFL parts are in sync...

PR #759, opened 2019-12-04
Change osRtxInfo init to avoid warning

kjbracey-arm:
GCC 9 produces a quite reasonable warning about the odd initialisation of `osRtxInfo`:

    [Warning] rtx_kernel.c@93,3: 'memset' offset [17, 164] from the
    object at 'osRtxInfo' is out of the bounds of referenced subobject
    'kernel' with type 'struct <anonymous>' at offset 8 [-Warray-bounds]

Change the initialisation to be expressed in terms of `&osRtxInfo` to avoid the warning - I feel this also makes the operation clearer.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_kernel.c
+++ b/CMSIS/RTOS2/RTX/Source/rtx_kernel.c
@@ -89,8 +89,8 @@ static osStatus_t svcRtxKernelInitialize (void) {
   }
 #endif
 
-  // Initialize osRtxInfo
-  memset(&osRtxInfo.kernel, 0, sizeof(osRtxInfo) - offsetof(osRtxInfo_t, kernel));
+  // Initialize osRtxInfo from kernel member onwards (leaving the OS ID and Version)
+  memset((char *)&osRtxInfo + offsetof(osRtxInfo_t, kernel), 0, sizeof(osRtxInfo) - offsetof(osRtxInfo_t, kernel));
 
   osRtxInfo.isr_queue.data = osRtxConfig.isr_queue.data;
   osRtxInfo.isr_queue.max  = osRtxConfig.isr_queue.max;


kjbracey-arm:
Actually, why is the memset even there? I was assuming it was to handle re-initialisation, but checking more closely, you don't support that anyway.

So, my revised suggestion - remove the memset.

RobertRostohar:
memset is there to initialize the osRtxInfo members that have not been explicitly initialized.

Can we be sure that members that are not specified in the initializing list will be initialized to zero? Regardless of compilers and user linker scripts?

kjbracey-arm:
> Can we be sure that members that are not specified in the initializing list will be initialized to zero? Regardless of compilers and user linker scripts?

Unnamed members in non-automatic variables being zero-init is part of the language. I assume we can trust the compiler, so the question is then linker/runtime.

If you can't trust an environment to initialise the unnamed members to zero, you also can't trust it to initialise the `os_id` and `version` fields.

As the `osRtxInfo` is partially initialised, the whole structure ends up in the `.data` section rather than `.bss`, so it's all subject to `memcpy` rather than `memset`. The zero fill is going to be part of the same `memcpy` that sets the ID and version. 

It would probably be a space optimisation, if the `os_id` and `version` fields were actually initialised by code in `osKernelInitialize`, so there's no need for ROM init data for `osRtxInfo`.  Then you'd be relying on proper `memset` of `.bss`. I think you're already relying on that in a few places, eg `osRtxTimerMemUsage` (assuming compiler sticks explicit all-zeros in bss - they do these days).

But some toolchains do compress their init data, so all those zeroes wouldn't take much space on them, and maybe you have a solid reason to want those fields initialised early.

kjbracey-arm:
I can see a precautionary argument to clear it, in case someone has managed to mess with it between boot and `osKernelInitialize`. But that would be very weird - all OS calls should be blocked by `osRtxKernelInactive` state. I can imagine checking that as an "assert-all-zeros", perhaps.

JonatanAntoni:

kjbracey-arm:
Rereading the discussion, this patch produces a lint warning instead. Although I believe the patch is valid and lint's warning is less reasonable than GCC's, that would need addressing?

So my alternative suggestion was to remove the memset altogether. I can make a patch for that if there's agreement.

JonatanAntoni:

RobertRostohar:
GCC 10 2020-q4 does not produce the warning anymore (looks as there was a bug in the offset calculation).

So is this still relevant?

kjbracey-arm:
Do you have a reference to the "bug"? The warning is accurate - this is zero-initting more than `osRtxInfo.kernel` - but maybe they've decided to tighten when they raise it based on some other conditions.

RobertRostohar:
Why is the warning accurate?

GCC 9 produces a warning:
```
rtx_kernel.c(93): warning: 'memset' offset [17, 164] from the object at 'osRtxInfo' is out of the bounds of referenced subobject 'kernel' with type 'struct <anonymous>' at offset 8 [-Warray-bounds]
```

But generates code which does not zero-init more then it should:
```
../../../Source/rtx_kernel.c **** 
 .loc 1 93 3 view .LVU59
 movs r2, #156
 add r0, r5, #8
 bl memset
```

kjbracey-arm:
As it says, you're taking the address of the `kernel` member, and setting off of its end. You're splatting the other members.

That's certainly worth warning about. Quite possible you got the size of `kernel` wrong.

The only time I think that would be a common idiom is in an array - something like `memset(&array[0], 0, 8)` where you obviously want to write to more than `array[0]`.

RobertRostohar:
OK, I see now. I was misreading the warning and only focusing on the generated code in regards to the whole osRtxInfo structure. Let me take another look.

kjbracey-arm:
As I said previously, at this stage I vote for just removing the `memset` altogether,

RobertRostohar:
The `memset` has been removed as suggested.

Thank you for you patience and sorry for not addressing this earlier.

IngmarPaetzold:
Hi there, I just had an equivalent "critical buffer overflow" by a different static analysis tool with RTX version 5.6.
My assumption is as well that the tool sees member .kernel and figures out that memset writes beyond its boundaries.

I am pretty sure - presumed that sizeof() and offsetof() are correctly working - that the memset is doing at least no harm. So can anyone confirm that the memset did what it was supposed to (i.e. also not zero-ing fields that it should _not_ to)?

(finding it being removed in the current master supports that, I guess...)

JonatanAntoni:
@IngmarPaetzold, I can confirm that the memset did exactly what it should do. The memset was used to zero all members of the struct starting at a certain offset. Just the way the statement was written the compiler and static analysis tools were not able to properly detect that. In the meanwhile the code in question has been rewritten.

PR #772, opened 2019-12-16
CMSIS-NN: Optimized mat_mul_s8 with DSP acceleration

gmiodice:
The new implementation does not require temporary buffers and assumes that the RHS input matrix is transposed (conv1x1 case)

### DIFF
--- a/CMSIS/Core/Include/cmsis_armcc.h
+++ b/CMSIS/Core/Include/cmsis_armcc.h
@@ -62,9 +62,9 @@
 #ifndef   __STATIC_INLINE
   #define __STATIC_INLINE                        static __inline
 #endif
-#ifndef   __STATIC_FORCEINLINE                 
+#ifndef   __STATIC_FORCEINLINE
   #define __STATIC_FORCEINLINE                   static __forceinline
-#endif           
+#endif
 #ifndef   __NO_RETURN
   #define __NO_RETURN                            __declspec(noreturn)
 #endif
@@ -460,7 +460,7 @@ __STATIC_INLINE void __set_FPSCR(uint32_t fpscr)
  */
 #define __DMB()                           __dmb(0xF)
 
-                  
+
 /**
   \brief   Reverse byte order (32 bit)
   \details Reverses the byte order in unsigned integer value. For example, 0x12345678 becomes 0x78563412.
@@ -874,6 +874,21 @@ __attribute__((always_inline)) __STATIC_INLINE uint32_t __USAT(int32_t val, uint
 
 #define __SMMLA(ARG1,ARG2,ARG3)          ( (int32_t)((((int64_t)(ARG1) * (ARG2)) + \
                                                       ((int64_t)(ARG3) << 32U)     ) >> 32U))
+
+__STATIC_FORCEINLINE uint32_t __SXTB16_ROR8(uint32_t x)
+{
+  return (__SXTB16(__ROR(x, 8)));
+}
+
+__STATIC_FORCEINLINE uint32_t __SXTB16_ROR16(uint32_t x)
+{
+  return (__SXTB16(__ROR(x, 16)));
+}
+
+__STATIC_FORCEINLINE uint32_t __SXTB16_ROR24(uint32_t x)
+{
+  return (__SXTB16(__ROR(x, 24)));
+}
 
 #endif /* ((defined (__ARM_ARCH_7EM__) && (__ARM_ARCH_7EM__ == 1))     ) */
 /*@} end of group CMSIS_SIMD_intrinsics */
--- a/CMSIS/Core/Include/cmsis_armclang.h
+++ b/CMSIS/Core/Include/cmsis_armclang.h
@@ -594,7 +594,7 @@ __STATIC_FORCEINLINE void __TZ_set_FAULTMASK_NS(uint32_t faultMask)
   Devices without ARMv8-M Main Extensions (i.e. Cortex-M23) lack the non-secure
   Stack Pointer Limit register hence zero is returned always in non-secure
   mode.
-  
+
   \details Returns the current value of the Process Stack Pointer Limit (PSPLIM).
   \return               PSPLIM Register value
  */
@@ -640,7 +640,7 @@ __STATIC_FORCEINLINE uint32_t __TZ_get_PSPLIM_NS(void)
   Devices without ARMv8-M Main Extensions (i.e. Cortex-M23) lack the non-secure
   Stack Pointer Limit register hence the write is silently ignored in non-secure
   mode.
-  
+
   \details Assigns the given value to the Process Stack Pointer Limit (PSPLIM).
   \param [in]    ProcStackPtrLimit  Process Stack Pointer Limit value to set
  */
@@ -1435,6 +1435,30 @@ __STATIC_FORCEINLINE int32_t __SMMLA (int32_t op1, int32_t op2, int32_t op3)
 
   __ASM volatile ("smmla %0, %1, %2, %3" : "=r" (result): "r"  (op1), "r" (op2), "r" (op3) );
   return(result);
+}
+
+__STATIC_FORCEINLINE int32_t __SXTB16_ROR8(int32_t op1)
+{
+  int32_t result;
+
+  __ASM volatile ("sxtb16 %0, %1, ROR #8" : "=r" (result) : "r" (op1) );
+  return (result);
+}
+
+__STATIC_FORCEINLINE int32_t __SXTB16_ROR16(int32_t op1)
+{
+  int32_t result;
+
+  __ASM volatile ("sxtb16 %0, %1, ROR #16" : "=r" (result) : "r" (op1) );
+  return (result);
+}
+
+__STATIC_FORCEINLINE int32_t __SXTB16_ROR24(int32_t op1)
+{
+  uint32_t result;
+
+  __ASM volatile ("sxtb16 %0, %1, ROR #24" : "=r" (result) : "r" (op1) );
+  return (result);
 }
 
 #endif /* (__ARM_FEATURE_DSP == 1) */
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -46,9 +46,9 @@
 #ifndef   __STATIC_INLINE
   #define __STATIC_INLINE                        static inline
 #endif
-#ifndef   __STATIC_FORCEINLINE                 
+#ifndef   __STATIC_FORCEINLINE
   #define __STATIC_FORCEINLINE                   __attribute__((always_inline)) static inline
-#endif                                           
+#endif
 #ifndef   __NO_RETURN
   #define __NO_RETURN                            __attribute__((__noreturn__))
 #endif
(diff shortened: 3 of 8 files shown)


CMSIS/DSP/Include/arm_math.h
```diff
@@ -357,7 +357,7 @@ extern "C"
 
 
 /* Included for instrinsics definitions */
-#if defined (_MSC_VER ) 
+#if defined (_MSC_VER )
```
JonatanAntoni: Well, doesn't directly correspond to this PR but should we discuss to move such compiler dependencies to a specific compiler_msc.h rather than cluttering the math header?

CMSIS/DSP/Include/arm_math.h
```diff
@@ -1676,6 +1676,42 @@ __STATIC_INLINE q31_t arm_div_q63_to_q31(q63_t num, q31_t den)
                        ((((q31_t)x <<  8) >>  8) & (q31_t)0xFFFF0000)  ));
   }
 
+   /*
+   * @brief C custom defined SXTB16 with immediate rotate right 8 bits
+   */
+  __STATIC_FORCEINLINE uint32_t __SXTB16_ROR8(
```
JonatanAntoni: Where shall those SXTB16_RORn function be located? In CMSIS compiler headers or in math header?

CMSIS/NN/Source/ConvolutionFunctions/arm_nn_mat_mult_s8.c
```diff
@@ -155,4 +155,384 @@ q7_t *arm_nn_mat_mult_s8(const q7_t *input_row,
...
+    (void)n;
+    (void)k;
+    (void)lhs_offset;
+    (void)dst_offset;
+    (void)activation_min;
+    (void)activation_max;
+    (void)dst;
+    return NULL;
+#endif
 }
```
JonatanAntoni: New line at end of file should be fixed.

CMSIS/Core/Include/cmsis_armcc.h
```diff
@@ -62,9 +62,9 @@
 #ifndef   __STATIC_INLINE
   #define __STATIC_INLINE                        static __inline
 #endif
-#ifndef   __STATIC_FORCEINLINE                 
+#ifndef   __STATIC_FORCEINLINE
```
JonatanAntoni: Mixing white space clean up with function enhancements makes it hard to see the relevant ones. Perhaps we spit this into two PRs? I can easily accept the cleanup if it doesn't contain function changes.

JonatanAntoni:
Would it be feasible to split this PR into 3, one per affected module, i.e.
- CMSIS-Core: Enhancements to the Compiler headers
- CMSIS-DSP: Enhancements to the Math header
- CMSIS-NN: Actual feature for NN

CMSIS/DSP/Include/arm_math.h
kjbracey-arm: I would say CMSIS compiler header.

ACLE doesn't have it, it just has `__sxtb16` and `__ror`, but that's documenting compiler extensions, and as a compiler extension it's reasonable to expect the compiler to be able to combine the two. And it seems that ARMC5 and ARMC6 at least do.

But if forced to use assembler on at least one compiler, so no chance of optimisation, then an having an explicit `__SXTB16_ROR` makes sense alongside `__SXTB16`.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1962,6 +1962,30 @@ __STATIC_FORCEINLINE uint32_t __SXTB16(uint32_t op1)
   return(result);
 }
 
+__STATIC_FORCEINLINE uint32_t __SXTB16_ROR8(uint32_t op1)
+{
+  uint32_t result;
+
+  __ASM volatile ("sxtb16 %0, %1, ROR #8" : "=r" (result) : "r" (op1) );
```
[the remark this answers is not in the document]
JonatanAntoni: Is using Assembler necessary on GCC? If such a (special) function is really needed I'd go for this generic as possible solution. Cluttering the Compiler headers with too many optimized stuff looks overwhelming.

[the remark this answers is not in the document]
kjbracey-arm: Assembler is necessary, both for basic SXTB16 and with-rotate.

If IAR also can optimise __SXTB16(__ROR()), then that leaves GCC as the only thing which would actually need something custom.

You could maybe put that conditional somewhere in the math header

    #if defined __GNUC__ && !defined __ARMCC_VERSION
    static inline __SXTB16_ROR assembler definition
    #else
    #define __SXTB16_ROR(op1, rotate) __SXTB(__ROR(op1, rotate))
    #endif

I guess you're right that lots of the __XXX opcode helpers in cmsis_compiler.h are limited to only basic "Rout,Rin" forms, missing extra op2 flexibility.

CMSIS/DSP/Include/arm_math.h
GuentherMartin: I would add such functions to CMSIS compiler header.

CMSIS/DSP/Include/arm_math.h
christophe0606: This one is very special and just to be able to build Python or Matlab wrapper. It is focusing on being able to compile for x86.

There is no reason to have support for this compiler in a separate file and it would not make sense to have such an header in Core or Core_A.

The problem is that there is no pure C version of all DSP functions. m0 versions are using 3 intrinsics which required a C version to be able to compile the Python or Matlab wrappers.

CMSIS/DSP/Include/arm_math.h
christophe0606: Here those functions are pure C fallback when no intrinsics are available.
In the compiler headers you have the corresponding versions with builtin or asm.

I don't think the pure C versions should move to the compiler headers.

Also, don't forget that now we also want to be able to compile CMSIS-DSP without using ANY of the compiler headers (for connecting to system like matlab or scipy which is really important for people designing signal processing code).

kjbracey-arm: Ah, these fallback ones shouldn't really be here at all. They should just be defined as __SXTB16(__ROR(op1, rotate)), just as for ARMC5 and ARMC6 (and IAR? which is missing from this PR?).

Everyone should be using that definition, really - only GCC (and maybe IAR) gains an optimisation by the 1-instruction assembly.

kjbracey-arm:
Flagging an issue a bit more loudly - this is missing a __SXTB16_ROR definition for a bunch of compilers, including IAR.

So we'd need even more boilerplate in this form.

What I think I'd suggest here, for this "optimisation" type instruction, is to do something like

    // ===cmsis_gcc.h===
    static inline uint32_t __CMSIS_GCC_SXTB16_ROR(...) asm...
    #define __SXTB16_ROR(op1, rotate)  __CMSIS_GCC_SXTB16_ROR(op1, rotate)
    // (or just do definition directly as macro like __SSAT?)
   
    // ===cmsis_compiler.h===
    #ifndef __SXTB16_ROR
    #define __SXTB16_ROR(op1, rotate) __SXTB16(__ROR(op1, rotate))
    #endif

I think I'd prefer code inside Core/Include relying on `__SXTB16_ROR` being a detectable macro to having code outside do that.

CMSIS/DSP/Include/arm_math.h
christophe0606: I don't see this one __SXTB16_ROR8 on my version of arm_math.h on develop branch.
For this one, I agree it can be built from other ones.

For other, I disagree. We have the strong requirement that we need to be able to compile CMSIS-DSP without any reference to any of the existing compiler headers.

The version of the functions as close to a pure C version as possible are the m0 version (ARM_MATH_DSP off). But even in this case, there are a few remaining intrinsics in use in the code and so a C implementation of them is required.

This C implementation could be in a different header but this header would have to be in DSP/Includes. 

But this discussion is not really relevant for this pull request.

kjbracey-arm: I am specifically talking about having custom definitions for `_SXTB16_ROR` that manually unroll it, rather than using `__SXTB16(__ROR())`. I'm not talking about the other fallbacks.

christophe0606: Ok. I agree.

CMSIS/DSP/Include/arm_math.h
JonatanAntoni: Being able to compile code on host machines (Windows, Linux) is not very special to DSP. One might want to do that for unit testing or perhaps for "simulating" target applications. Not sure what's actually worth to have it as part of CMSIS.

JonatanAntoni:
cc: @TTornblom please cross check the Compiler header enhancement from IAR point of view.

CMSIS/DSP/Include/arm_math.h
christophe0606: And where do you want to have it ? People are getting CMSIS-DSP as part of CMSIS. They want to be able to compile it for host machines. So, the code has to support it.

kjbracey-arm: From my point of view, it would be helpful if Core had C fallbacks for its intrinsics.

Mbed OS uses Core + RTOS without DSP, and we have to worry about unit tests fallbacks for things like `__RBIT` ourselves.

christophe0606: If we can have a cleaner solution answering the needs of everybody then I am all for it.

As usual, it is a matter of trade-off. We did the minimal fix in CMSIS-DSP to enable this feature (host compilation). It has been very well received by users of the library. 

Doing it cleanly is more work and not just impacting CMSIS-DSP.

kjbracey-arm: This is going off on a slight tangent, but as of GCC 9, it's possible to have a "best-of-both-worlds" for the cases where assembler is required for run-time speed.

You can use a C version for constant evaluation, and the assembler only for runtime, meaning the compiler can constant-evaluate things like `__RBIT(0)`. This previously was only possible when __RBIT was built-in.

See https://github.com/ARMmbed/mbed-os/pull/11897

CMSIS/Core/Include/cmsis_armclang.h
```diff
@@ -1437,6 +1437,30 @@ __STATIC_FORCEINLINE int32_t __SMMLA (int32_t op1, int32_t op2, int32_t op3)
   return(result);
 }
 
+__STATIC_FORCEINLINE int32_t __SXTB16_ROR8(int32_t op1)
+{
+  int32_t result;
+
+  __ASM volatile ("sxtb16 %0, %1, ROR #8" : "=r" (result) : "r" (op1) );
```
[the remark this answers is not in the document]
gmiodice: I agree. I am going to remove this assembler for armclang as suggested

gmiodice:
Hi @JonatanAntoni,

I totally agree to split this PR in 3 parts. I am going to abandon this PR and create 3 different PRs as suggested by your comment.
Thanks.

gmiodice:
As requested, I created a different PR for the SXTB16_ROR support. (https://github.com/ARM-software/CMSIS_5/pull/795/files)

felix-johnny:
@gmiodice .. Can we close this PR?

PR #775, opened 2019-12-18
CMSIS-NN: Fixed bug in arm_q7_to_q15_with_offset

gmiodice:
The issue was related to the offset casted to int8_t rather than int16_t.
In case of int8_t offset, we could not handle properly the cases with negative offset.
In particular, the creation of the vector variable offset_q15x2 was not correct
since we did not extend the sign bit

Signed-off-by: Gian Marco Iodice <gianmarco.iodice@arm.com>

### DIFF
--- a/CMSIS/NN/Include/arm_nnsupportfunctions.h
+++ b/CMSIS/NN/Include/arm_nnsupportfunctions.h
@@ -133,7 +133,7 @@ void      arm_q7_to_q15_reordered_no_shift(const q7_t * pSrc, q15_t * pDst, uint
  * </pre>
  *
  */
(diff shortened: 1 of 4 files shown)


JonatanAntoni:
Just a general note: Please be careful when changing APIs. In this case: Is moving from `q7_t` to `q15_t` backward compatible? If not you break existing user code. That means you would need to increase the CMSIS-NN major version to 2.

felix-johnny:
@JonatanAntoni ok.. This is an internal API(not part of arm_nnfunctions.h) and backward compatible. 

@gmiodice Could you make a patch that updates the major version in the changed files to 2.?

gmiodice:
Sure I can. I create the PR to update the major version.

PR #779, opened 2019-12-19
arm_add_q7: rewrite while-loop to for-loop. NFC.

sjoerdmeijer:
Non-functional change to rewrite this counting down while-loop into a for-loop,
and also using array indexing instead of pointer arithmetic.

This is a little experiment how you guys feel about simplifying this code: I
hope it's easy to agree that first of all this increases readability of the
code and, believe it or not, makes the life of the compiler and
auto-vectorisation easier. I am in the process of fixing the vectoriser who is
struggling with this counting down loop (in the context of MVE and
tail-predication), but thought that a rewrite of this loop would be good too.

If we like this, I think I will give a few more kernels a similar treatment.

### DIFF
--- a/CMSIS/DSP/Source/BasicMathFunctions/arm_add_q7.c
+++ b/CMSIS/DSP/Source/BasicMathFunctions/arm_add_q7.c
@@ -140,15 +140,10 @@ void arm_add_q7(
 
 #endif /* #if defined (ARM_MATH_LOOPUNROLL) */
 
-  while (blkCnt > 0U)
+  for (int i = 0; i < blkCnt; i++)
   {
     /* C = A + B */
-
-    /* Add and store result in destination buffer. */
-    *pDst++ = (q7_t) __SSAT((q15_t) *pSrcA++ + *pSrcB++, 8);
-
-    /* Decrement loop counter */
-    blkCnt--;
+    pDst[i] = (q7_t) __SSAT((q15_t) pSrcA[i] + pSrcB[i], 8);
   }
 
 }


GuentherMartin:
Hello @sjoerdmeijer,
the DSP function use such kind of loop very often.
Do all the loops need to be rewritten to generate more effective code?

sjoerdmeijer:
Hi Martin,

> the DSP function use such kind of loop very often.

Yep, indeed, and that's what I would like to change first for the so called "Cortex-M0" loops, i.e. the reference implementation without any manual optimisations applied. We are benchmarking the compiler optimisations with these "cortex-m0" loops against the hand-tuned kernels. That's the context why I like these reference loops as clean and compiler-friendly as possible.

These cortex-m0 reference kernels are actually the best and most easy input to analyse for the compiler. The hand-tuned kernels interfere with the compiler optimisations. And that's exactly the point I am trying to address: when compilers and their transformations were not yet so powerful as they are today, it was often more performant to explicitly and hand tuned code, like manual loop unrolling and pointer-arithmetic, so that the compiler can do a straightforward translation. But today we have powerful loop-transformations and auto-vectotisation, and exactly these old DSP habits of writing code are biting us now. 

There's a whole lot of programming literature on this, but arrays are not pointers. Designating objects as arrays and not as pointers help the compiler and its loop transformations.
This is pointer arithmetic: *A++. And this is array indexing: A[i], although this is a bit of syntactic sugar decaying to the same address calculation. However, since the generated code is the same, and A[i] is more pleasant to read than *A++, I think the former is preferred. Under the compiler hood, the intermediate code is different though, just illustrating that the compiler has to work a bit harder. 
A bit of topic, but also illustrating that arrays are not pointers, It would be good to put restrict on the pointers of the arguments, which indicates that this is not just an arbitrary pointer that can point to anything, but the compiler can assume that memory of these arguments allowing more transformations. The compiler can deal with this, it will emit runtime checks to make and create different code path, so the performance will be the same, but just another illustration the compiler has to work harder, and in this case there will be one negative effect: a slight code bloat for this extra emitted code.

I think the for-loop is also preferred over the counting down while-loop. With the for-loop all expression are on 1 line, but with the while-loop there is the decrement i-- statement at the end of the loop. 
For the compiler, however, an incrementing for-loop is the more canonical form that loop transformations expects loops to be in. I am addressing that for the vectoriser, but as I said, I think this is better to have for-loops.

GuentherMartin:
Hello @sjoerdmeijer,
Thank you for the detailed explanation.
A deeper knowledge of how the compiler works helps to choose the appropriate C instructions to get the best optimization results.

JonatanAntoni:

christophe0606:
Those changes will be applied progressively since they impact lot of functions.
So, I keep this pull request open and tagged as future to remember the action.

christophe0606:
It will be tracked in this new github issue on the new repository : https://github.com/ARM-software/CMSIS-DSP/issues/24

PR #780, opened 2019-12-19
Removed unnecessary casts

lamarrr:
Casting using **void*** gives a compiler error (on _arm-none-eabi-g++_), as it discards the **const qualifier**.
**void*** cast is not needed since **read_and_pad_reordered**'s first parameter is **const q7_t***
Also, **read_and_pad_reordered** returns **const q7_t***, so the **q7_t*** cast is not needed since variable **pB** is already **const q7_t***.

### DIFF
--- a/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_connected_q7.c
+++ b/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_connected_q7.c
@@ -146,7 +146,7 @@ arm_fully_connected_q7(const q7_t * pV,
         {
             q31_t     inV1, inV2, inM11, inM12;
 
-            pB = (q7_t *) read_and_pad_reordered((void *)pB, &inM11, &inM12);
+            pB = read_and_pad_reordered(pB, &inM11, &inM12);
 
             inV1 = *__SIMD32(pA)++;
             sum = __SMLAD(inV1, inM11, sum);


lamarrr:
> What exactly is the compiler saying? I don't see any actual error in the code, just bad style.
> 
> Unusually strict warning level?

An -fpermissive error. It's an error to cast away const qualifiers.

Remember to use these compiler flags: -Wall -Wextra -Wpedantic

kjbracey-arm:
> An -fpermissive error. It's an error to cast away const qualifiers.

It's undefined behaviour to modify a `const` object by casting away const qualifiers. It is not an error to merely cast a qualifier away, and in this case it happens to be totally safe and legal because you immediately assign the non-const-qualified pointers to const-qualified ones again.

But, yes, diagnosing the cast makes sense as a "pedantic" warning - it's a dangerous thing to do that leads to real errors.

But remember the reason it's called `-Wpedantic` is that it gives diagnostics about things that are NOT (necessarily) invalid code.

lamarrr:
> > An -fpermissive error. It's an error to cast away const qualifiers.
> 
> It's undefined behaviour to modify a `const` object by casting away const qualifiers. It is not an error to merely cast a qualifier away, and in this case it happens to be totally safe and legal because you immediately assign the non-const-qualified pointers to const-qualified ones again.
> 
> But, yes, diagnosing the cast makes sense as a "pedantic" warning - it's a dangerous thing to do that leads to real errors.
> 
> But remember the reason it's called `-Wpedantic` is that it gives diagnostics about things that are NOT (necessarily) invalid code.

Yeah, I understand that. The warnings help prevent unsafe and liable logical flaws. 
It's better staying on the safer side. 
I'm using arm-none-eabi-g++ to compile (compiles C code too, though as C++. Ignorable if you're not targeting compatibility with C++ codebases).
It's valid C but not valid C++, irregardless of the pedantic or any of the flags as it doesn't pass the compilation stage.

Moreso, there's no reason to cast there.
A **const void*** and const **q7_t***  cast might be better and safer. 
The function signature is:
_const q7_t* read_and_pad_reordered(const q7_t *..._

I feel this is used by other C++ projects like Tensorflow Lite Micro. Google uses the 4 warning flags in their projects.

kjbracey-arm:
Agree there's absolutely no reason to cast, and I'm usually the first person to point that out in code, which is why I immediately approved it.

I'm just being picky and trying to pin down what the compiler's telling you and why (because your report doesn't say).

Yes, it isn't valid C++, but that's still nothing to do with the `const`. That would be down to the implicit `void *` to `q7_t *` conversion.

So is that the actual error you're seeing?

(I thought it was `-Wcast-qual`, but I don't think that's enabled by any of the options you've shown. It really would be easier if you'd just paste the compiler diagnostic!)

lamarrr:
> Agree there's absolutely no reason to cast, and I'm usually the first person to point that out in code, which is why I immediately approved it.
> 
> I'm just being picky and trying to pin down what the compiler's telling you and why (because your report doesn't say).
> 
> Yes, it isn't valid C++, but that's still nothing to do with the `const`. That would be down to the implicit `void *` to `q7_t *` conversion.
> 
> So is that the actual error you're seeing?
> 
> (I thought it was `-Wcast-qual`, but I don't think that's enabled by any of the options you've shown. It really would be easier if you'd just paste the compiler diagnostic!)


My bad, really sorry

Here you go:

```
CMSIS/CMSIS_5/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_co
nnected_q7.c: In function 'arm_status arm_fully_connected_q7(const q7_t*, const q7_t*, uint16_t, uint16_t, uint16_t, uint16_t, const q7_t*, q7_t*, q15_t*)':
CMSIS/CMSIS_5/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_connected_q7.c:149:49: error: invalid conversion from 'void*' to 'const q7_t*' {aka 'const signed char*'} [-fpermissive]
  149 |             pB = (q7_t*) read_and_pad_reordered((void*)pB, &inM11, &inM12);
      |                                                 ^~~~~~~~~
      |                                                 |
      |                                                 void*
In file included from CMSIS/CMSIS_5/CMSIS/NN/Include/arm_nnfunctions.h:118,
                 from CMSIS/CMSIS_5/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_connected_q7.c:32:
CMSIS/CMSIS_5/CMSIS/NN/Include/arm_nnsupportfunctions.h:320:69: note:   initializing argument 1 of 'const q7_t* read_and_pad_reordered(const q7_t*, q31_t*, q31_t*)'
  320 | __STATIC_FORCEINLINE const q7_t *read_and_pad_reordered(const q7_t *source, q31_t * out1, q31_t * out2)
      |                                                         ~~~~~~~~~~~~^~~~~~
```

kjbracey-arm:
Great, thanks. Yes, it's the `void *` to `[const] q7_t *` assignment that's the problem. Illegal C++ with or without the `const`.

Could you adjust the Git commit message and GitHub description, to lose the incorrect `const` bits? I'd say just "remove unnecessary casts" is sufficient self-justification, but feel free to note that it makes the file C++ compatible.

lamarrr:
> Great, thanks. Yes, it's the `void *` to `[const] q7_t *` assignment that's the problem. Illegal C++ with or without the `const`.
> 
> Could you adjust the Git commit message and GitHub description, to lose the incorrect `const` bits? I'd say just "remove unnecessary casts" is sufficient self-justification, but feel free to note that it makes the file C++ compatible.

Alright, will do. Thanks!

lamarrr:
> Great, thanks. Yes, it's the `void *` to `[const] q7_t *` assignment that's the problem. Illegal C++ with or without the `const`.
> 
> Could you adjust the Git commit message and GitHub description, to lose the incorrect `const` bits? I'd say just "remove unnecessary casts" is sufficient self-justification, but feel free to note that it makes the file C++ compatible.

Probably not the best place for this.
I want to implement some NN kernels (BasicMath). I'm only familiar with 32-bit SIMD and not MVEI.
How do I go about implementing it?

PR #787, opened 2019-12-29
implemented negate kernel for s8

lamarrr:
Implemented the SIMD equivalent of TensorFlow lite micro's negate kernel. 

However, I don't have access to an HELIUM-compatible MCU, so I can't implement its kernel. I thus, made it fall back to the default un-optimized version in case SIMD is not supported by the device.

### DIFF
--- a/CMSIS/NN/Include/arm_nnfunctions.h
+++ b/CMSIS/NN/Include/arm_nnfunctions.h
@@ -1371,6 +1371,19 @@ extern    "C"
                                     const int32_t out_activation_min,
                                     const int32_t out_activation_max,
                                     const uint32_t block_size);
+
+   /**
+   * @brief s8 element-wise negation of a vector
+   * @param[in, out]       vect            pointer to input vector
+   * @param[in]            size            number of elements
+   * @return none
+   * 
+   * @details   Supported framework: TensorFlow Lite micro
+   */
+
+    void      arm_negate_s8(int8_t *vect, uint16_t size);
+
+
 /**
  * @defgroup Acti Activation Functions
  *
--- a/CMSIS/NN/Source/BasicMathFunctions/arm_negate_s8.c
+++ b/CMSIS/NN/Source/BasicMathFunctions/arm_negate_s8.c
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2010-2019 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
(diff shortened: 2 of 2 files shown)


felix-johnny:
@lamarrr Thanks for the contributions. could you give the motivation(mainly network where you see this) for the need of an optimized version for the negate function? Are you using it from TFLite Micro and if so made a PR for the glue/wrapper functions there?

lamarrr:
I can't think of any neural network that uses it. I found it under TfLite Micro and  I thought CMSIS-NN is meant to serve as a backend to provide optimized implementations for all kernels found under Tensorflow Lite Micro (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/kernels/neg.cc)

felix-johnny:
@lamarrr For simple ones like this, i would recommend  to use the C reference versions in TFLite Micro instead. So, I'll close this PR. 
 If you are interested in developing/improving heavily used kernels, please let me know and I can point out a few. Thanks again!

lamarrr:
Yeah, though this is far faster than the reference version.
I am interested and would appreciate you pointing them out, thanks!

felix-johnny:
@lamarrr That is great. One API that is needed is a SIMD optimized version for depthwise conv when the channel multiplier is not 1. Right now the pure C version is used(https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/kernels/cmsis-nn/depthwise_conv.cc). I would suggest to make a version for the case where the depth multipler is multiple of 4 or 8 and not a generic one. The API would look the same as the one for the optimized version(depth_mult = 1).
  If you would like to do this, let me know so I can make a note and we don't do a double effort on it.

felix-johnny:
@lamarrr And the use of this in the person detect model that you can find in TFL Micro.

lamarrr:
> @lamarrr And the use of this in the person detect model that you can find in TFL Micro.

Great! 
I'd love to. But I don't ha

> @lamarrr That is great. One API that is needed is a SIMD optimized version for depthwise conv when the channel multiplier is not 1. Right now the pure C version is used(https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/kernels/cmsis-nn/depthwise_conv.cc). I would suggest to make a version for the case where the depth multipler is multiple of 4 or 8 and not a generic one. The API would look the same as the one for the optimized version(depth_mult = 1).
> If you would like to do this, let me know so I can make a note and we don't do a double effort on it.

Can I get a full list of the yet-to-be-optimized kernels?

felix-johnny:
@lamarrr the message was not complete. The rest are further improvements to existing functions.

PR #795, opened 2020-01-10
CMSIS CORE/DSP: Adding support for __SXTB16_ROR

gmiodice:


CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1927,14 +1927,14 @@ __STATIC_FORCEINLINE uint32_t __USADA8(uint32_t op1, uint32_t op2, uint32_t op3)
 #define __SSAT16(ARG1, ARG2) \
 ({                          \
   int32_t __RES, __ARG1 = (ARG1); \
-  __ASM volatile ("ssat16 %0, %1, %2" : "=r" (__RES) :  "I" (ARG2), "r" (__ARG1) : "cc" ); \
+  __ASM ("ssat16 %0, %1, %2" : "=r" (__RES) :  "I" (ARG2), "r" (__ARG1) : "cc" ); \
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1927,14 +1927,14 @@ __STATIC_FORCEINLINE uint32_t __USADA8(uint32_t op1, uint32_t op2, uint32_t op3)
...
   __RES; \
  })
 
 #define __USAT16(ARG1, ARG2) \
 ({                          \
   uint32_t __RES, __ARG1 = (ARG1); \
-  __ASM volatile ("usat16 %0, %1, %2" : "=r" (__RES) :  "I" (ARG2), "r" (__ARG1) : "cc" ); \
+  __ASM ("usat16 %0, %1, %2" : "=r" (__RES) :  "I" (ARG2), "r" (__ARG1) : "cc" ); \
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1974,31 +1974,31 @@ __STATIC_FORCEINLINE uint32_t __SMUAD  (uint32_t op1, uint32_t op2)
 {
   uint32_t result;
 
-  __ASM volatile ("smuad %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
+  __ASM ("smuad %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1974,31 +1974,31 @@ __STATIC_FORCEINLINE uint32_t __SMUAD  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE uint32_t __SMUADX (uint32_t op1, uint32_t op2)
 {
   uint32_t result;
 
-  __ASM volatile ("smuadx %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
+  __ASM ("smuadx %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1974,31 +1974,31 @@ __STATIC_FORCEINLINE uint32_t __SMUAD  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE uint32_t __SMLAD (uint32_t op1, uint32_t op2, uint32_t op3)
 {
   uint32_t result;
 
-  __ASM volatile ("smlad %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
+  __ASM ("smlad %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -1974,31 +1974,31 @@ __STATIC_FORCEINLINE uint32_t __SMUAD  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE uint32_t __SMLADX (uint32_t op1, uint32_t op2, uint32_t op3)
 {
   uint32_t result;
 
-  __ASM volatile ("smladx %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
+  __ASM ("smladx %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -2040,31 +2040,31 @@ __STATIC_FORCEINLINE uint32_t __SMUSD  (uint32_t op1, uint32_t op2)
...
+  __ASM ("smusdx %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
   return(result);
 }
 
 __STATIC_FORCEINLINE uint32_t __SMLSD (uint32_t op1, uint32_t op2, uint32_t op3)
 {
   uint32_t result;
 
-  __ASM volatile ("smlsd %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
+  __ASM ("smlsd %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -2040,31 +2040,31 @@ __STATIC_FORCEINLINE uint32_t __SMUSD  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE uint32_t __SMLSDX (uint32_t op1, uint32_t op2, uint32_t op3)
 {
   uint32_t result;
 
-  __ASM volatile ("smlsdx %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
+  __ASM ("smlsdx %0, %1, %2, %3" : "=r" (result) : "r" (op1), "r" (op2), "r" (op3) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -2106,23 +2106,23 @@ __STATIC_FORCEINLINE uint32_t __SEL  (uint32_t op1, uint32_t op2)
 {
   uint32_t result;
 
-  __ASM volatile ("sel %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
+  __ASM ("sel %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
```
kjbracey-arm: This needs the volatile - it sets the GE flags.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -2106,23 +2106,23 @@ __STATIC_FORCEINLINE uint32_t __SEL  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE  int32_t __QADD( int32_t op1,  int32_t op2)
 {
   int32_t result;
 
-  __ASM volatile ("qadd %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
+  __ASM ("qadd %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -2106,23 +2106,23 @@ __STATIC_FORCEINLINE uint32_t __SEL  (uint32_t op1, uint32_t op2)
...
   return(result);
 }
 
 __STATIC_FORCEINLINE  int32_t __QSUB( int32_t op1,  int32_t op2)
 {
   int32_t result;
 
-  __ASM volatile ("qsub %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
+  __ASM ("qsub %0, %1, %2" : "=r" (result) : "r" (op1), "r" (op2) );
```
kjbracey-arm: This needs the volatile - it sets Q.

CMSIS/DSP/Include/arm_math.h
```diff
@@ -1746,6 +1746,24 @@ __STATIC_INLINE q31_t arm_div_q63_to_q31(q63_t num, q31_t den)
...
 #endif /* !defined (ARM_MATH_DSP) */
 
+  /*
+   * @brief Function to extract two 8-bit values from an operand (op1) rotated by 8/16/24 bits (rotate) and sign-extend them to 16 bits each
+   * @param[in] op1     input operand
+   * @param[in] rotate  number of bits (allowed 8/16/24) to rotate the input operand
+   * @return            the 8-bit values sign-extended to 16-bit values
+   */
+#if defined __GNUC__ && !defined __ARMCC_VERSION && defined(ARM_MATH_DSP)
+  __STATIC_FORCEINLINE uint32_t __SXTB16_ROR(uint32_t op1, uint32_t rotate)
```
kjbracey-arm: I'd prefer this to always be an inline function, if only for Doxygen purposes, with the condition inside its definition.

What's the `ARM_MATH_DSP` test for?

christophe0606: This must not be in arm_math.h. In arm_math.h we have (for historical reasons - will be cleaned in future) some pure C implementation of a few intrinsics (when ARM_MATH_DSP is not defined). But that's all .

gmiodice:
Hi @kjbracey-arm ,
as suggested I remove the volatile changes in this review.

CMSIS/DSP/Include/arm_math.h
gmiodice: ARM_MATH_DSP has been added because I noticed a fall back for SXTB16 in case of ARM_MATH_DSP not defined. In the same file there is the C custom defined SXTB16 when ARM_MATH_DSP is not passed at compile time. 

@christophe0606 , where could I place this new function?

christophe0606: @gmiodice Normally the support for new intrinsics and the asm code is going into the compiler files in Core and CoreA.

If you need also a pure C fallback, and because we currently don't have another solution, you can put the pure C fallback  in arm_math.h

But, as much as possible we also try to avoid this. When pure C fallback is required we provide a pure C function of the algorithms in DSP.

gmiodice: Hi @christophe0606 , in my previous PR it was suggested to put it here (https://github.com/ARM-software/CMSIS_5/pull/772#discussion_r359284523). Do you suggest that it is worth to have this code in cmsis_gcc.h for instance?

christophe0606: @gmiodice What is clear is that it cannot be in arm_math.h. It is not the goal of this header nor the direction where we want to go in the future to clean this header (which is really needed).

Now, cmsis_gcc.h looks like a natural place to put this (but I may not know all of your constraints).

gmiodice: many thanks @christophe0606 for the clarification. The main problem is that if I write this function in cmsis_gcc, I will need to update all the compiler headers as well otherwise the function cannot be compiled in all cases. I was wondering if I could write this new function directly in my "NN" function although I will need it for more than one function as well

christophe0606: @gmiodice If it is very specific to NN and you don't anticipate other uses in short term, then perhaps better to keep it in NN for now which will avoid all those changes to compiler headers. And if in the future, we see that this change may benefit other codes than NN, then it will be time to update compiler headers.

But, in any case, you'll need to ensure it builds with the compilers supported by CMSIS.

kjbracey-arm: > When pure C fallback is required we provide a pure C function of the algorithms in DSP.

Note that the "pure C fallback" is the same as the "optimised ARM assembler version" for everything except GCC - just use `__SXTB16(__ROR()))`. So I'd be inclined to put that in "cmsis_compiler.h".

Maybe like:

    forceinline uint32_t __SXTB16_ROR(...)
    {
    #ifdef __COMPILER_SXTB16_ROR // cmsis_gcc.h can define this as its assembler version
        return __COMPILER_SXTB16_ROR(...);
    #else
        return __SXTB16(__ROR(...));
    #endif
    }

kjbracey-arm: > in my previous PR it was suggested to put it here

Yes, that was my first suggestion - I later reconsidered. See https://github.com/ARM-software/CMSIS_5/pull/772#discussion_r359307822

gmiodice: Hi @kjbracey-arm , my bad. I missed the last suggestion from the discussion. From the snipped above means that __COMPILER_SXTB16_ROR has to be passed at compile time in order to trigger the assembler routine defined in cmsis_gcc.h for SXTB16_ROR. 

So the mbed build command will be something like:

mbed compile -m auto -t GCC_ARM -f -DARM_MATH_LOOPUNROLL -DARM_MATH_DSP -D__COMPILER_SXTB16_ROR

Note: Should I call it ARM_SXTB16_ROR instead of __COMPILER_SXTB16_ROR to keep the convection used so far?

kjbracey-arm: I mean that cmsis_gcc.h itself would define `__COMPILER_SXTB16_ROR`. Nothing required from the user.

Being a macro serves as an indicator to cmsis_compiler.h that the toolchain-specific version exists, and that macro is itself the definition.

It could be an all-in-one definition as one of those funny GCC extended block expressions like `__SSAT` is or could be defined to invoke another static inline function (`__gcc_SXTB16_ROR` or something).

Don't have any strong feelings on naming.

PR #799, opened 2020-01-15
CMSIS-NN: Improve performance of 1x1 conv for MVE

felix-johnny:
Performance of 1x1 conv is improved for Multi Vector Extension.
Added two new files:
1. arm_nn_mat_mul_core_1x_s8.c
2. arm_nn_mat_mul_core_4x_s8.c

### DIFF
--- a/CMSIS/NN/Source/NNSupportFunctions/arm_nn_mat_mul_core_4x_s8.c
+++ b/CMSIS/NN/Source/NNSupportFunctions/arm_nn_mat_mul_core_4x_s8.c
@@ -1,0 +1,117 @@
+/*
+ * Copyright (C) 2010-2020 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


CMSIS/NN/Source/NNSupportFunctions/arm_nn_mat_mul_core_4x_s8.c
```diff
@@ -0,0 +1,117 @@
...
+    int32_t acc_n2 = 0;
+    int32_t acc_n3 = 0;
+
+    const int8_t *ip_row_0 = row_base;
+    const int8_t *ip_row_1 = row_base + row_elements;
+    const int8_t *ip_row_2 = row_base + (2 * row_elements);
+    const int8_t *ip_row_3 = row_base + (3 * row_elements);
+    int32_t sum_tmp = 0;
+
+#if defined(ARM_MATH_MVEI)
```
christophe0606: You probably need another flag for the future since it will have to also compile on other compilers. And I am not sure the embedded asm syntax is always the same for all compilers.

So, I think it would be safer to have another compilation flag to select the ASM version or the intrinsic version.

felix-johnny: @christophe0606 all right.. i'll add one in the next patch..

PR #808, opened 2020-01-22
CMSIS-NN: 1xN conv support for MVE

felix-johnny:
1. MVE specific function for 1xN convolution.
2. Updated use of macro ARM_MATH_AUTOVECTORIZE
3. Updated history for new function that is added.

### DIFF
--- a/CMSIS/NN/Include/arm_nnfunctions.h
+++ b/CMSIS/NN/Include/arm_nnfunctions.h
@@ -78,12 +78,15 @@
    *
    * - ARM_MATH_DSP:
    *
-   * Define macro ARM_MATH_DSP, If the silicon supports DSP instructions.
+   * Define macro ARM_MATH_DSP, If the silicon supports DSP instructions(DSP extension).
    *
    * - ARM_MATH_MVEI:
    *
    * Define macro ARM_MATH_MVEI, If the silicon supports M-Profile Vector Extension.
-   *
+
+   * - ARM_MATH_AUTOVECTORIZE
+   *  Used in conjucture with ARM_MATH_MVEI to let the compiler auto vectorize for the functions that uses inline assembly.
+   *  It does not affect functions that use C or intrinsics.
    * - ARM_MATH_BIG_ENDIAN:
    *
    * Define macro ARM_MATH_BIG_ENDIAN to build the library for big endian targets. This is supported only for the legacy functions i.e, functions targetted at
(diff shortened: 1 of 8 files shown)


CMSIS/NN/Source/NNSupportFunctions/arm_nn_mat_mul_core_4x_s8.c
```diff
@@ -57,40 +58,40 @@ arm_status arm_nn_mat_mul_core_4x_s8(const int32_t row_elements,
...
-           ,[row1] "+r"(ip_row_1)
-           ,[row2] "+r"(ip_row_2)
-           ,[row3] "+r"(ip_row_3)
-           ,[out0] "+Te"(acc_n0)
-           ,[out1] "+Te"(acc_n1)
-           ,[out2] "+Te"(acc_n2)
-           ,[out3] "+Te"(acc_n3)
-           :[cnt] "r"(row_elements)
-           :"q0","q1", "q2","q3","q4", "memory", "r14");
+#if defined(ARM_MATH_MVEI) && !defined(ARM_MATH_AUTOVECTORIZE)
```
christophe0606: I still think another flag is needed to select C or Asm (in case this Asm specific syntax is not supported on a given C compiler).
It is not the goal of ARM_MATH_AUTOVECTORIZE flag which can also be applied to C code.
ARM_MATH_AUTOVECTORIZE is used to select a version of the code which is working well with auto vectorizer (so no unrolling, not too many intrisinsics ...)

CMSIS/NN/Include/arm_nnfunctions.h
```diff
@@ -78,12 +78,15 @@
...
    *
-   * Define macro ARM_MATH_DSP, If the silicon supports DSP instructions.
+   * Define macro ARM_MATH_DSP, If the silicon supports DSP instructions(DSP extension).
    *
    * - ARM_MATH_MVEI:
    *
    * Define macro ARM_MATH_MVEI, If the silicon supports M-Profile Vector Extension.
-   *
+
+   * - ARM_MATH_AUTOVECTORIZE
```
christophe0606: This flag is not related to ASM.
I need to document the meaning of this flag in arm_math.h
But today it is not yet applied very consistently in the DSP library.

CMSIS/NN/Source/NNSupportFunctions/arm_nn_mat_mul_core_4x_s8.c
felix-johnny: @christophe0606 ok. i'll create a new one..

PR #825, opened 2020-02-11
CMSIS-NN Opt: Optimize Depthwise convolution 3x3 w/ unit multiplier o…

giorgio-arenarm:
…n Cortex-M4/M7 (non-DSP)

### DIFF
--- a/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_3x3_s8.c
+++ b/CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_3x3_s8.c
@@ -1,0 +1,209 @@
+/*
+ * Copyright (C) 2010-2020 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_s8_opt.c
```diff
@@ -22,7 +22,7 @@
  * Description:  Optimized s8 depthwise separable convolution function for
  *               channel multiplier of 1.
  *
- * $Date:        January 21, 2020
+ * $Date:        February 2020
```
felix-johnny: Looks good with the performance improvements on this. Lets make it as its own function(arm_depthwise_conv_3x3_s8()) to not mix the use of additional buffers. I can see it being used as it is for the MVE case with just just a modification for the requantization. 

When that is done, please add it to the TFLu glue as well under a defined(ARM_MATH_DSP) && !defined(ARM_METH_MVEI).

giorgio-arenarm: Done. Will push to TFLu once this is merged

CMSIS/NN/Source/ConvolutionFunctions/arm_depthwise_conv_3x3_s8.c
```diff
@@ -0,0 +1,209 @@
...
+ * limitations under the License.
+ */
+
+/* ----------------------------------------------------------------------
+ * Project:      CMSIS NN Library
+ * Title:        arm_depthwise_conv_3x3_s8.c
+ * Description:  Optimized s8 depthwise convolution function for channel
+ *               multiplier of 1 and 3x3 kernel size.
+ *
+ * $Date:        February 2020
```
felix-johnny: please specify a date as well.. same for arm_nnfunctions.h .. rest looks good..

PR #842, opened 2020-03-02
CMSIS-NN: Improve performance of convolution for MVE

felix-johnny:
Further opttimization is done for the regular convolution case
for Multi-Vector Extention case.

### DIFF
--- a/CMSIS/NN/Include/arm_nnsupportfunctions.h
+++ b/CMSIS/NN/Include/arm_nnsupportfunctions.h
@@ -21,8 +21,8 @@
  * Title:        arm_nnsupportfunctions.h
  * Description:  Public header file of support functions for CMSIS NN Library
  *
- * $Date:        7 February 2020
- * $Revision:    V.3.0.0
+ * $Date:        March 1, 2020
+ * $Revision:    V.4.0.0
  *
  * Target Processor:  Cortex-M cores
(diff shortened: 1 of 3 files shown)


CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -633,6 +633,36 @@ __STATIC_FORCEINLINE q31_t arm_nn_requantize(const q31_t val, const q31_t multip
                                        RIGHT_SHIFT(shift));
 }
 
+/**
```
christophe0606: Why not use the DSP/Support/arm_copy_q7 ?
The DSP version is not using predication. Not sure it will make a big difference (or perhaps only for small buffers).

felix-johnny: @christophe0606 Ah.. didn't know that there was a version there! I had tried a similar non-predicated one the version that you see here. There were some minor differences in the performance, but the major difference was in in-lining. It is replaced by 4 instructions. Nice and compact.  If arm_copy_q7() can be made as an inline function, I am happy to use that.

christophe0606: @felix-johnny  Ok but it will have to wait for the big rework of arm_math.h

Issue #617, opened 2019-06-25
[cmsis_gcc.h] nested extern declaration


salkinium:
When compiling with `-Wnested-externs`, these warnings appear:

```
modm/ext/cmsis/core/cmsis_gcc.h: In function '__cmsis_start':
modm/ext/cmsis/core/cmsis_gcc.h:133:15: warning: nested extern declaration of '_start' [-Wnested-externs]
   extern void _start(void) __NO_RETURN;
               ^~~~~~
modm/ext/cmsis/core/cmsis_gcc.h:146:31: warning: nested extern declaration of '__copy_table_start__' [-Wnested-externs]
   extern const __copy_table_t __copy_table_start__;
                               ^~~~~~~~~~~~~~~~~~~~
modm/ext/cmsis/core/cmsis_gcc.h:147:31: warning: nested extern declaration of '__copy_table_end__' [-Wnested-externs]
   extern const __copy_table_t __copy_table_end__;
                               ^~~~~~~~~~~~~~~~~~
modm/ext/cmsis/core/cmsis_gcc.h:148:31: warning: nested extern declaration of '__zero_table_start__' [-Wnested-externs]
   extern const __zero_table_t __zero_table_start__;
                               ^~~~~~~~~~~~~~~~~~~~
modm/ext/cmsis/core/cmsis_gcc.h:149:31: warning: nested extern declaration of '__zero_table_end__' [-Wnested-externs]
   extern const __zero_table_t __zero_table_end__;
                               ^~~~~~~~~~~~~~~~~~
```

I'm currently working around this by `#define __PROGRAM_START 1`, but that's pretty brute force.

salkinium:
Uses the latest version of these files.

JonatanAntoni:
Hi @salkinium,

thanks for pointing this out. These warnings are caused by the reworked C startup code (compiler agnostic). I'll check what's behind and how we can change it. 

As long as you are using the former startup routines (compiler specific) you are fine with defining `__PROGRAM_START` yourself. The new startup code would fail if `__PROGRAM_START` doesn't name a proper library (or application) start routine.

Cheers,
Jonatan

salkinium:
Yeah, redefining `__PROGRAM_START`is very hacky and fragile. Unfortunately, since it‘s a header file, it‘s a bit difficult to not leak the extern variables everywhere… So I understand why they were extern declared inside the function.

JonatanAntoni:
Hi @salkinium,

[...]

PR #854, opened 2020-03-06
Issue #617, compiling error fix in local structure declarations.

noxuz:
Avoids global scope of the structures, as discussed in https://github.com/ARM-software/CMSIS_5/issues/617

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -132,22 +132,17 @@ __STATIC_FORCEINLINE __NO_RETURN void __cmsis_start(void)
 {
   extern void _start(void) __NO_RETURN;
   
-  typedef struct {
+  struct __copy_table_t{
     uint32_t const* src;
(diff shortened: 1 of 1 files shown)


CMSIS/Core/Include/cmsis_gcc.h
```diff
@@ -132,22 +132,17 @@ __STATIC_FORCEINLINE __NO_RETURN void __cmsis_start(void)
 {
   extern void _start(void) __NO_RETURN;
   
-  typedef struct {
+  struct __copy_table_t{
     uint32_t const* src;
     uint32_t* dest;
     uint32_t  wlen;
-  } __copy_table_t;
+  } __copy_table_start__, __copy_table_end__;
```
JonatanAntoni: This defines new symbols in local scope. The code below relies on getting these symbols defined by the linker. Hence the symbols were declared as defined externally.

noxuz: I appreciate the clarification.

PR #899, opened 2020-04-19
arm_max_pool_s8.c: use quotes instead of <> for arm_math.h include

sandeepmistry:
This makes the `arm_math.h` include consistent with the rest of the files in CMSIS-NN, where `#include "arm_math.h` is used instead of `#include <arm_math.h>`.


felix-johnny:
@sandeepmistry Thanks!

PR #904, opened 2020-04-23
CMSIS-NN: CMSIS-NN structs for the TensorFlowLite micro compliant functions

gmiodice:
…ctions

### DIFF
--- a/CMSIS/NN/Include/arm_nn_types.h
+++ b/CMSIS/NN/Include/arm_nn_types.h
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2020 Arm Limited or its affiliates. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


LaurentF06:
Sounds good.
I would move input_offset and output_offset to "cmsis_nn_per_channel_params" structure, and maybe rename this one "cmsis_nn_quantization_params" to remain generic.

gmiodice:
Hi @LaurentF06 ,

many thanks for your valuable comment/feedback. 

The reason why we decided to adopt a struct called cmsis_nn_per_channel_params rather than a single and generic one is only to avoid a possible confusion on the quantization scheme adopted by the functions.

The idea is to have two different structs for the quantization:

cmsis_nn_per_channel_quant_params
cmsis_nn_per_tensor_quant_params

At the moment in the PR there is only the first one and with the wrong name (it does not include _quant_). Based on your name suggestion, I will update the PR this afternoon.

The reason why we would like to have two different quantization schemes is to address the two quantization schemes currently needed for convolution layer (per channel quantization) and fully connected layer (per tensor quantization).

Of course, as suggested we could potentially put together in a unique struct but, in case of a future support for per channel quantization in fully connected layer, (we think) this could cause a confusion on the quantization scheme adopted. 

Basically with these two structs we would have:

# Per channel quantization
fully_connected_s8(cmsis_nn_per_channel_quant_params quant, ...)
fully_connected_s8(cmsis_nn_per_tensor_quant_params quant, ...)

In this manner the developer knows exactly what parameters of the quantization are needed and required to be initialized properly.

Regarding your consideration to move input_offset and output_offset in the quantization struct, we decided to follow the same approach of TFL/TFLu where the input offset and output offset are part of convolution parameters whilst the multiplier and shift part of the quantization.

At the following link https://www.tensorflow.org/lite/performance/quantization_spec#specification_summary you can find the specification for per channel and per tensor quantization. As described, the quantization only refers to scale and the zero points which are actually translated to multiplier and shift in the implementation.

Hope this can help. 

Once again many thanks for your comment and do not hesitate to write if you have further comments/considerations/suggestions

Gian Marco

PR #905, opened 2020-04-28
CMSIS-NN: Add new unit test cases for Convolution

felix-johnny:
1. 5 new unit test cases are added targetting convolution.
2. Format of stride passed on in generating reference data
   is fixed.


CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_s8/test_arm_convolve_s8.c
```diff
@@ -104,3 +108,225 @@ void stride2pad1_arm_convolve_s8(void)
...
+{
+  arm_status expected = ARM_MATH_SIZE_MISMATCH;
+  q7_t output[CONV_1_X_N_1_DST_SIZE] = {0};
+
+  const int32_t buf_size = arm_convolve_1_x_n_s8_get_buffer_size(CONV_1_X_N_1_IN_CH,
+                                                                 CONV_1_X_N_1_FILTER_X,
+                                                                 CONV_1_X_N_1_FILTER_Y);
+  q15_t *bufferA = (q15_t*)malloc(buf_size);
+
+  arm_status result = arm_convolve_1_x_n_s8(conv_1_x_n_1_input,
```
mansnils, line 245: arm_convolve_1_x_n_s8 should perhaps be in its own file?
TestCases/test_arm_convolve_1_x_n_s8/test_arm_convolve_1_x_n_s8.c

felix-johnny: Atleast for the DSP extension(which is the target here), it uses arm_convolve_s8 at the backend. So,  went with the same file..

mansnils: Ok, let's go for that

PR #908, opened 2020-04-29
CMSIS-NN: Use arithmetic shifting in arm_nn_mult

### LINKED ISSUE #891: arm_cfft_init_f64.c not listed in CMakeLists.txt
I have noticed that the `arm_cfft_init_f64.c` is not listed in the `CMakeLists.txt`. Is this intentional?

Also the `arm_cfft_init_f64` function is not covered by the test suite -- the current `TransformCF64.cpp` implementation relies solely on the static tables. Is this also intentional?

https://github.com/ARM-software/CMSIS_5/blob/93b52c8a563140ba236b439c3466c4044c404641/CMSIS/DSP/Testing/Source/Tests/TransformCF64.cpp#L44

### LINKED ISSUE #896: negative value is not shifted correclty
As NN_ROUND is defined as `1u<<shift_value`, it's an unsigned type. It causes product of src and dst (see code below) is logical shifted rather than arithmatic shift.

https://github.com/ARM-software/CMSIS_5/blob/d82d3cc52a2ccccafa55fbc84a6fbc59e5f85531/CMSIS/NN/Source/NNSupportFunctions/arm_nn_mult_q7.c#L79

I use arm compiler v6.10.1, cortex-m33

arvoelke:
Fixes #896.

I have not tested this directly. Let me know if a unit test is needed (I see some examples/instructions in the [UnitTest](https://github.com/ARM-software/CMSIS_5/tree/develop/CMSIS/NN/Tests/UnitTest) directory that I could follow). I've tested it indirectly through an example in the issue:

```C
#include <stdio.h>

int main() {
  int x = -32;
  printf("%d\n", (x >> 1));  // arithmetic shift
  printf("%d\n", (((int) x + 0u) >> 1));  // logical shift (becomes unsigned before shift)
  printf("%d\n", ((int) (x + 0u) >> 1));  // arithmetic shift (cast back to signed before shift)
  return 0;
}
```

Using `gcc` (version 7.5.0) this outputs:
```
-16
2147483632
-16
```

The middle line corresponds to the old code. The last line corresponds to the updated code. 

One question:

1. (see commit in PR)
```
(q15_t) __SSAT(((q31_t) ((q31_t) (*pSrcA++) * (*pSrcB++) + NN_ROUND(out_shift)) >> out_shift), 16);
```

2. (alternative)
```
(q15_t) __SSAT((((q31_t) (*pSrcA++) * (*pSrcB++) + (q31_t) NN_ROUND(out_shift)) >> out_shift), 16);
```

Is the second option better in any way? I think it is equivalent, and maybe clearer?

### DIFF
--- a/CMSIS/NN/Source/NNSupportFunctions/arm_nn_mult_q15.c
+++ b/CMSIS/NN/Source/NNSupportFunctions/arm_nn_mult_q15.c
@@ -21,8 +21,8 @@
  * Title:        arm_nn_mult_q15.c
  * Description:  Q15 vector multiplication with variable output shifts
  *
- * $Date:        13. July 2018
- * $Revision:    V.1.0.0
+ * $Date:        29. April 2020
+ * $Revision:    V.1.0.1
  *
  * Target Processor:  Cortex-M cores
  *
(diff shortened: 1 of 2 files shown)


arvoelke:
By the way, I did a search through the repo for uses of `NN_ROUND` and briefly checked them all. They all looked fine to me.

felix-johnny:
@arvoelke Option 2 is clearer..but option 1 is right as well. Thanks for checking and for the contribution!

PR #909, opened 2020-05-02
CMSIS-NN: Fix batch dimension in arm_convolve_s8

Tombana:
This fixes the way batches are handled in arm_convolve_s8.c used by TF Lite Micro.

See also https://github.com/ARM-software/CMSIS_5/pull/708#discussion_r417083087


felix-johnny:
@Tombana Thanks for the contribution!

PR #917, opened 2020-05-11
CMSIS-NN: Fix and add conv TC for batch > 1

felix-johnny:
1. Incorrect output reference array size in case of batch !=1
    is fixed for the TC reference
2. TC conv_4 TC is added
3. conv_1_x_n_1 is re generated

TC - Test Case


felix-johnny:
@mansnils  Could you review this?

CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_s8/test_arm_convolve_s8.c
```diff
@@ -190,6 +191,51 @@ void conv_3_arm_convolve_s8(void)
 
   free(bufferA);
 }
+#if 1
```
mansnils, line 194: Forgot to remove #if 1/#endif?

felix-johnny: yep! will fix it.

PR #918, opened 2020-05-12
CMSIS-NN: Fix warning in arm_nnsupportfunctions.h

felix-johnny:
Warnings related to sign conversions are looked at
and typecasted.


felix-johnny:
@christophe0606 True.. This is just scratching the surface..but a start :-)

PR #923, opened 2020-05-19
CMSIS-NN: Adopted the new API interface in convolution functions

gmiodice:
The following functions have been ported to use the new API interface:
- arm_convolve_1_x_n_s8
- arm_convolve_1x1_s8_fast
- arm_convolve_s8

A wrapper function has been also created to make the integration of
those functions easier

PR #924, opened 2020-05-19
CMSIS-NN: Update unit test for convolution as per new interface

felix-johnny:
The API's are updated to reflect the modified interface.


felix-johnny:
Dependent on https://github.com/ARM-software/CMSIS_5/pull/923

Issue #925, opened 2020-05-21
DelayStarted and DelayUntilStarted missing thread_id


joelbyren:
I'm trying to limit the number of events to Event Recorder, so I'm only running with "Op" events on. This will sometimes make it unclear which thread is performing DelayStarted or DelayUntilStarted, since they always arrive when a thread is switched out.

I suggest adding the thread_id of the referenced thread in these events, which should not affect performance too much since there is already one unused parameter for them.

PR #926, opened 2020-05-21
Suggested solution to #925 (DelayStarted and DelayUntilStarted event update)

joelbyren:
Here is a suggestion on updating DelayStarted and DelayUntilStarted to have them reference the thread they are referring to, since they occur after the thread is switched out.

Not sure if pull request should include updated version number of RTX5.scvd, but it is included.


JonatanAntoni:

joelbyren:
Let me know if you need any further updates; I have not checked the code since I submitted the Pull Request, so I hope it still works.

JonatanAntoni:

PR #933, opened 2020-05-27
define ARM_WIFI_HOSTNAME

MRT-77:
define ARM_WIFI_HOSTNAME for use with ARM_WIFI_SetOption and ARM_WIFI_GetOption in CMSIS-Driver


JonatanAntoni:
Hi MRT-77,

thanks for proposing this enhancement.

Can you please describe the use case for this new option? Which driver implementation shall make use of it? In which way shall this option be used?

Cheers,
Jonatan

MRT-77:
Hi

One of the features of Network Component is changing the Hostname, and I found that this feature does not exist in the new WiFi driver.
I added this macro for use in the CMSIS-Driver pack and tested it in Keil software.

[CMSIS-Driver Pull request](https://github.com/ARM-software/CMSIS-Driver/pull/5)

for example:
```
ARM_DRIVER_WIFI *wifi = &Driver_WiFi0;
// Set
wifi->SetOption(0, ARM_WIFI_HOSTNAME, "Test", 4);

uint32_t len = 32;
char buf[len];
// Get
wifi->GetOption(0, ARM_WIFI_HOSTNAME, buf, &len);
```

MRT-77:
The latest change is to use **BSSID**.
If you have multiple WiFi with a similar name, you can use BSSID to set your target.

JonatanAntoni:

PR #939, opened 2020-06-02
Fix - arm_math.h: Use the float suffix in arm_clarke_f32

M-Bab:
The float suffix replaces the need for a float cast in arm_clarke_f32.
This change is necessary to avoid large amount of warnings by arm_math.h when the -Wunsuffixed-float-constants flag is used. And using this compile flag is very reasonable on a processor with 32bit FPU.


JonatanAntoni:
Hi @M-Bab,

thanks for pointing this out.

Cheers,
Jonatan

PR #946, opened 2020-06-19
Initial Cortex-A AArch64 support.

alfedotov:
Novadays there are a lot of MCU's with A53 onboard. Admittedly this "Application profile" core and intended to be used with some OS like Linux etc. On another hand there are applications that doesn't require OS presence but demands more performance than Cortex-M offers.
I think it would nice to have unified API for AArch64 as well.
Comments, updates and fixes are welcomed.

### DIFF
--- a/CMSIS/Core_AArch64/Include/cmsis_compiler.h
+++ b/CMSIS/Core_AArch64/Include/cmsis_compiler.h
@@ -1,0 +1,245 @@
+/**************************************************************************//**
+ * @file     cmsis_compiler.h
+ * @brief    CMSIS compiler generic header file
+ * @version  V1.0.0
+ * @date     15. June 2020
+ ******************************************************************************/
+/*
+ * Copyright (c) 2020 Arm Limited. All rights reserved.
+ *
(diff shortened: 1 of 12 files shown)


alfedotov:
No comments from anyone ?
New Cortex-R82 should go there as well since it is AArch64 only...

JonatanAntoni:

CMSIS/Core_AArch64/Include/cmsis_compiler.h
```diff
@@ -0,0 +1,245 @@
+/**************************************************************************//**
+ * @file     cmsis_compiler.h
```
rpmsousa, line 2: Hi,

Did you consider adding the support inside the existing Core_A directory? (to reduce duplication)

alfedotov: Well, yes I've considered to use Core_A directory at the very beginning. But then I've ran into the problem that I have to pollute existing headers by AArch64 routine implementations.

PR #949, opened 2020-06-24
Core(M): #undef IAR standard definition of __WEAK

TTornblom:
The IAR standard definition of __WEAK is incompatible with the
normal CMSIS definition. This fix #undef:s this definition


JonatanAntoni:
Merged as 614cb83.

PR #952, opened 2020-06-26
CMSIS-DSP: Optimize Q15 radix4 FFT for GCC with PKHBT for Cortex-M4

rsaxvc:
Update the radix4 Q15 FFTs to explicitly use the __PKHBT macros, and update the GCC header to implement it.

This makes Q15 FFT execution time a few percent shorter depending on FFT size when targeting Cortex-M4.


christophe0606:
Changed merged as detailed in https://github.com/ARM-software/CMSIS_5/issues/968

Issue #962, opened 2020-07-09
CMSIS-DAP: firmware version in DAP_Info


flit:
The DAP_Info command specification for the `DAP_ID_FW_VER` identifier simply states the value is "CMSIS-DAP Firmware Version". The reference CMSIS-DAP implementation currently returns either "1.2.0" or "2.0.0", depending on the `DAP_FW_V1` macro. This implies that `DAP_ID_FW_VER` is, in fact, the CMSIS-DAP protocol version. Another way to phrase this is that it's the version of the reference implementation upon which the firmware is built.

It would be very helpful to more clearly specify the format and usage of the `DAP_ID_FW_VER` value. In addition, we should add a value that is clearly specified as the "firmware implementation version" rather than protocol version.

The context here is the request from the microbit foundation in ARMmbed/DAPLink#723 to be able to access the DAPLink firmware version from the host (so they can inform users that they should update the firmware). The latest code in DAPLink reports its own version for `DAP_ID_FW_VER`. So instead of a value like "2.0.0", you'll get back something like "0254". That solves the request, but it's unfortunate that you cannot also determine the CMSIS-DAP protocol version, and it's not clear whether it's compliant with the protocol.

RobertRostohar:
You are right. The CMSIS-DAP Firmware Version as defined in the DAP_Info command could be phrased as the CMSIS-DAP protocol version.

The version is taken into account by the debuggers in order to identify features of CMSIS-DAP.

Revision history shows the versions and how they evolved:
http://www.keil.com/pack/doc/CMSIS/DAP/html/dap_revisionHistory.html

Reporting any other version would not be compliant and debuggers might not work correctly.

Firmware implementation versions and how they are reported are device specific. They could be reported through the CMSIS-DAP Vendor Commands which could be also used for firmware updates (not part of CMSIS-DAP).

flit:
Needing get the implementation version is a very common requirement. If for nothing else other than presenting to the user. So we should allocate a standard info identifier for the firmware version. What about ID 0x07 (just the next in sequence)? I would also rename the constants in the code to be more clear…

Proposal:
- Change 0x04
  - Constant in code: `DAP_ID_FW_VER` ⟹ `DAP_ID_CMSIS_DAP_VER`
    - Will keep `DAP_ID_FW_VER` available as an alias of `DAP_ID_CMSIS_DAP_VER`, but deprecated
[...]

PR #973, opened 2020-07-27
CMSIS-DAP: DAP_Info documentation clarifications

flit:
These changes implement the proposal from #962, improving the version info reported by the DAP_Info command.

- Add ID 0x07, `DAP_ID_PRODUCT_VER`: **Product Firmware Version** (string), vendor-specified version format.
    - A stub for `DAP_ProductFirmwareVerString()` was added to `DAP_config.h`.
- Change ID 0x04 to be specified as: **CMSIS-DAP Protocol Version** (string). The value must be one of the versions from the \ref dap_revisionHistory "CMSIS-DAP revision history", such as "2.0.0"
    - Changed the constant in `DAP.h` to `DAP_ID_CMSIS_DAP_VER`.
    - Kept `DAP_ID_FW_VER` as an alias of `DAP_ID_CMSIS_DAP_VER`, with a comment saying it's deprecated.

Other included changes:
- Rebuilt the example projects to update the included .axf files.
- Added a `.gitignore` to ignore the generated `Objects` and `Listings` folders under the examples.
- Bumped CMSIS-DAP to version 2.1.0.

All changes have been tested in DAPLink and pyOCD.

### DIFF
--- a/CMSIS/DoxyGen/DAP/src/dap_USB_cmds.txt
+++ b/CMSIS/DoxyGen/DAP/src/dap_USB_cmds.txt
@@ -7,11 +7,11 @@ Commands between Debug Unit and host computer.
 This section explains each command that is exchanged between the Debug Unit and the host computer.
 Every Command starts with a Command-ID and optional data.
 Depending on the Command, the CMSIS-DAP firmware replies with a Response that repeats the Command-ID and
-delivers additional data. 
+delivers additional data.
 
 Command and Response data have a package size limitation that is defined with \ref DAP_PACKET_SIZE.
 This configuration setting can be obtained with the command \ref DAP_Info and is used to
-optimize the performance for Full-Speed or High-Speed USB. The debugger must ensure that each 
+optimize the performance for Full-Speed or High-Speed USB. The debugger must ensure that each
 data package fits within the limitations of the configured \ref DAP_PACKET_SIZE.
 
 \note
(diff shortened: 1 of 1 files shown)


flit:
cc @RobertRostohar

flit:
@RobertRostohar @JonatanAntoni Any chance we can move forward with this change? We unfortunately did not get it merged in time for inclusion in the micro:bit v2 firmware, but I would like to go ahead and fix it in the current DAPLink code base.

JonatanAntoni:

flit:
@JonatanAntoni Question about the string encoding for DAP_Info: is there any reason this cannot be changed to UTF-8? Some boards (in particular) have names that require accents or other non-ASCII glyphs. I can include an encoding change with this PR.

RobertRostohar:
We have already added the requested "Product Firmware Version". We also kept "CMSIS-DAP Firmware Version" because "Firmware" term is consistently used through the whole documentation (including in the name). Therefore renaming "Firmware" to "Protocol" dost not really fit.

flit:
Can you please take a look at the docs improvements in this PR? I've removed the other changes.

Regarding "Firmware" vs "Protocol", I'd be happy to clean up the rest of the documentation to be consistent with a change to "Protocol".

What about the ASCII -> UTF-8 encoding change?

Thanks!

flit:
Another question: can we change the description of "Vendor ID" and "Product ID" to "Vendor Name" and "Product Name"? Calling them "ID" is misleading because it's not the USB VID & PID, but the vendor/product name strings.

RobertRostohar:
@flit, thanks for all the improvements. They have now been merged together with a separate patch (mentioned above) with further improvements of the documentation for target strings.

The description of “Vendor ID” and “Product ID’ were misleading as you have pointed out. They are now renamed to use “Name” instead of “ID” (addressed in the mentioned patch).

RobertRostohar:
Regarding UTF-8 encoding for strings:

Strings could already be encoded with UTF-8 instead of ASCII. The Firmware would not need to change as there should not be any difference in handling.

However the debugger on the host side needs to know that it expects UTF-8 encoding and show the strings accordingly.

Until know we had just the Vendor Name, Product Name and Serial Number that would be effected. And even those where typically not provided directly but retrieved from USB information.

Therefore we could change the spec and state UTF-8 encoding instead of ASCII. Now would be probably a good time since we are introducing new strings.

However debuggers on host side would need to be adopted, otherwise they could show strange characters for non ASCII encoding.

RobertRostohar:
The string encoding has been changed to UTF-8 as discussed earlier.

PR #976, opened 2020-07-31
CMSIS-NN : Optimize requantization

felix-johnny:
The optmization takes into consideration
the ranges of arguments for requantization.

This affects almost all convolution,
depthwise convolution, fully connected APIs

### DIFF
--- a/CMSIS/DoxyGen/NN/src/history.txt
+++ b/CMSIS/DoxyGen/NN/src/history.txt
@@ -9,10 +9,10 @@
     <tr>
     <td>V1.x.0</td>
     <td>
-      Added wrapper functions for convolution related APIs
       <ul>
         <li>arm_depthwise_conv_wrapper_s8</li>
         <li>arm_convolve_wrapper_s8</li>
+        <li>arm_nn_doubling_high_mult_no_sat</li>
       </ul>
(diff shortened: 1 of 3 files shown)


CMSIS/DoxyGen/NN/src/history.txt
```diff
@@ -9,10 +9,10 @@
     <tr>
     <td>V1.x.0</td>
     <td>
-      Added wrapper functions for convolution related APIs
       <ul>
         <li>arm_depthwise_conv_wrapper_s8</li>
         <li>arm_convolve_wrapper_s8</li>
+        <li>arm_nn_sat_doubling_high_mult_no_sat</li>
```
dominicsymes, line 15: I'm just wondering if the name could be confusing containing both "sat" and "no_sat". Suggest just "arm_nn_doubling_high_mult()" or "arm_nn_doubling_high_mult_no_sat()"

felix-johnny: Thanks for the comment. I have gone with arm_nn_doubling_high_mult() for the saturating version and arm_nn_doubling_high_mult_no_sat() for the non saturating version. And I changed the names for the mve versions as well so it is uniform.

PR #991, opened 2020-09-04
CMSIS-NN: Fix output diff issue in unit test

mansnils:
- Set output scale to 1. Issue was that this could be slightly smaller
  than 1 depending on the input range, which could case output to diff
  against reference. Based on this some test data were updated.
- Increase input range in conv_2 test data set.
- Add option to run (generate) all test data sets at once.
- Update README with instructions on how to add a new test data set.


CMSIS/NN/Tests/UnitTest/generate_test_data.py
```diff
@@ -736,73 +746,111 @@ def generate_data(self, input_data=None, weights=None, biases=None):
         self.write_c_header_wrapper()
 
 
+def load_all_testdatasets():
+    """
+    Add all new testdata sets here
+    """
+
+    type_of_test = 'conv'
```
felix-johnny, line 754: This is much better!

felix-johnny (approved):

Issue #228, opened 2017-08-17
Round robin issue


gzzi:
Hi,

I have a issue that a task is never executed. I was able to reproduce it with 3 task:

 - TaskHighPrio: which use 50% of cpu ressources
 - TaskRoundRobin1: use as much time as possible
 - TaskRoundRobin2: use as much time as possible

The problem is that on rtx_system.c on function osRtxTick_Handler the robin.tick is reseted because half the time TickHandler come during TaskHighPrio and other half it come on TaskRoundRobin1. Then TaskRoundRobin2 is never started & executed.

I think round robin must be handled for each priority not just sampled on a tick basis.

JonatanAntoni:
Hi @AmberAussie,

Thank you for your report and the analysis. I was able to create a small reproducer that reveals the problem.

Currently we do not store round robin per priority. Its some kind of weighing up accuracy versus memory requirements.

An easy solution might be to put blocked threads behind all others of same priority instead of ahead. This would give all threads the chance to run but the actual time slices might not be equally distributed among all threads. I.e. if a round robin thread is preempted after using only part of its time slice the rest of the time slice is gone. Thus the other robin thread might actually get more execution time if it is less often preempted.

Do you have any concerns about such a "lazy" solution? A more elaborate solution would come at additional costs.

Kind regards,
Jonatan

AmberAussie:
Hi @JonatanAntoni ,

Thank you for your proposal, yes it would fix that particular problem but after reflection the fact that CPU is not be equally distributed is also a unexpected behavior on some other applications.

To achieve the correct scheduling, I don't see another solution than having a counter per priority or per thread or adding limitation (round robin between lowest priority task only or OS user shouldn't wake up of task < round robin time or ...)

I propose to add a robin_tick on os_thread_t that we initiate with osRtxInfo.thread.robin.timeout, decrements on osRtxTick_Handler until it reach 0. Once it reach 0 if there is another task with same priority ready we put the task back in thread list at the end of it's priority and reload robin_tick with osRtxInfo.thread.robin.timeout.

Or maybe something based on tick count between context switch... but it could be that it's often less than a tick so it's harder.

Kind regards

JonatanAntoni:
Thanks for sharing your thoughts. I will discuss the issue with the dev team and report back.

JonatanAntoni:
Hi @AmberAussie,

We discussed the issue in the team and apologize that the current implementation has some drawbacks.

[...]

PR #1000, opened 2020-09-23
fix thread insertion in ready list when blocked by higher priority thread

singds:
Hi,
I encountered the following issue using rtx.
There is a situation in which i ended up with one thread not sharing time with threads with same priority, in round-robin scheduling.
This happens when the thread in question is frequently interrupted by a higher priority thread.
In this interrupt the thread is blocked and reinserted in the ready list before other processes with same priority.
In the end, if a thread is always interrupted before it ends its timeslice, it never gives other threads with same priority the chance to run.
I think this can easily be fixed inserting the blocked thread in the ready list after al threads with same priority already in.

### DIFF
--- a/CMSIS/RTOS2/RTX/Source/rtx_thread.c
+++ b/CMSIS/RTOS2/RTX/Source/rtx_thread.c
@@ -400,7 +400,7 @@ static void osRtxThreadBlock (os_thread_t *thread) {
   prev = osRtxThreadObject(&osRtxInfo.thread.ready);
   next = prev->thread_next;
 
-  while ((next != NULL) && (next->priority > priority)) {
+  while ((next != NULL) && (next->priority >= priority)) {
     prev = next;
     next = next->thread_next;
   }


JonatanAntoni:
Hi @singds,

this is a known limitation already discussed in #228. 
While your patch resolves your immediate problem it has certain other drawbacks. We need to clearly think if the drawbacks caused by this minimal change might be acceptable.

Given the current limitation we recommend not mixing round robin and priority based scheduling.

Cheers,
Jonatan

singds:
Thank you for your rapid feedback,
Sorry i didn't notice the older issue.
I think the "fast" solution of putting blocked thread in the back of the ready list can be ok.
Maybe there are cases in which they don't equally share time but at least they share. This compromise is even easy to document.
The solution of a per thread counter is even better. I can try to implement this and update my pull request.
I think it's worth resolve the issue and i'm really happy if i can help.

About not mixing round robin with priority may be the case to write some line in the official documentation, because it shows actually an example of priority mixed with round-robin.
https://www.keil.com/pack/doc/CMSIS/RTOS2/html/theory_of_operation.html

JonatanAntoni:
Good catch. 

Changing the implementation, even if its such a small change, might affect existing applications. Basically, I think you change has a chance to work and deliver some sort of advantage. But as we might inappropriately affect existing application I tend to be *very* conservative.

Cheers,
Jonatan

singds:
I totally agree with you. The change is too delicate to be silently included.
Maybe we can add a config (default disabled) that enables it. This config could also be hidden from the _Configuration Wizard_ in RTX_Config.h if you don't like it to be so evident. What do you think ?
I wrote another solution with per thread robin counter. May be a little bit more fair although not perfect. I plan to test and submit it tomorrow, so you can give me your feedback.

JonatanAntoni:
Yes, probably we need to have some sort of config options here. We are due to enhance the scheduler, anyhow. This will take its time but we are happy to incorporate your proposals.

singds:
Hi,
I'm not happy with my solution with per thread counter.
There are still borderline cases i can not remove without modifying too much the code.
Moreover i think there is no need to overcomplicate, the "fast" way it's enough and do the job.
I'll open a case in keil to see if it helps speed the adoption of any fix.

Thanks

JonatanAntoni:

RobertRostohar:
Should be properly addressed with the patch ada4cde. Pull request #1000 should not be needed.

PR #1014, opened 2020-10-12
CMSIS-NN: Clean-up of include files

felix-johnny:
1. The wider arm_math.h is replaced by specific arm_math_types.h
   or arm_math_memory.h(one usage) files.

2. Only the necessary includes are added for all files. For most
   parts this has meant removing arm_nnsupportfunctions.h or
   arm_nnfunctions.h where it is not necessary. arm_math*.h is
   only a part of the include files and is removed from the source
   files.

The unit tests are not modified and have references to arm_math.h

### DIFF
--- a/CMSIS/NN/Include/arm_nn_types.h
+++ b/CMSIS/NN/Include/arm_nn_types.h
@@ -22,8 +22,8 @@
  * Description:  Public header file to contain the CMSIS-NN structs for the
  *               TensorFlowLite micro compliant functions
  *
- * $Date:        April 23, 2020
- * $Revision:    V.0.5.0
+ * $Date:        09. October 2020
+ * $Revision:    V.1.0.0
  *
  * Target Processor:  Cortex-M cores
  * -------------------------------------------------------------------- */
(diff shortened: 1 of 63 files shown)


felix-johnny:
@christophe0606 It is too many files! It is mainly to check the 4 files inside the include folder. Rest, I'll ask @freddan80 to have a look.

CMSIS/NN/Include/arm_nn_types.h
```diff
@@ -32,6 +32,8 @@
 #ifndef _ARM_NN_TYPES_H
 #define _ARM_NN_TYPES_H
 
+#include <stdint.h>
```
christophe0606, line 35: Why not include arm_math_types.h ? which is including cmsis_compiler.h which is including stdint.h

felix-johnny: I started off with that, but changed it as the file didn't include any vector register types and thought I'd include only what is necessary for this building block of a file. That was the reasoning behind not using arm_math_types.h here. I don't have a strong opinion in this case.

freddan80 (approved):

PR #1041, opened 2020-11-13
Do not build f16 code if datatype is disabled

LnnrtS:
to not build float16 code if the datatype is disabled.

Not sure if the condition should be extended. Feel free to adapt


christophe0606:
Thanks for fixing this.

PR #1062, opened 2020-12-03
CMSIS-NN: Fix fully connected unit test

mansnils:
Change quantization during generation of output reference data.

Change-Id: I6694324b42499763cf654d3a47ecb05962cac6d8


felix-johnny:
@mansnils  Please update the heading and description to reflect that is a bug fix. Atleast for the heading , maybe Fix instead of Update?

mansnils:
Ok

PR #1069, opened 2020-12-11
CMSIS-NN: Add new test cases for Fully Connected

felix-johnny:
Add new test cases to increase code coverage.

Change-Id: Idbebe59fcb94718ca24a9cc56473eb98a3cc4d3c


felix-johnny:
@mansnils Could you review this PR?

CMSIS/NN/Tests/UnitTest/generate_test_data.py
```diff
@@ -786,7 +786,15 @@ def load_all_testdatasets():
                                                         w_x=2, w_y=1, randmin=-4, randmax=4, batches=3,
                                                         input_zero_point=-50, weights_zero_point=-22,
                                                         output_zero_point=-2)
-
+    dataset = 'fully_connected_mve_0'
+    ALL_TESTDATA_SETS[dataset] = FullyConnectedSettings(dataset, type_of_test, args, in_ch=16, out_ch=9, x_in=1, y_in=1,
+                                                        input_zero_point=-3, w_x=1, w_y=1, randmin=-4, randmax=4, batches=1,
+                                                        output_zero_point=-2, outminrange=-125.0, outmaxrange=127.0)
+    dataset = 'fully_connected_mve_1'
```
mansnils, line 793: I realize now that outminrange/outmaxrange will produce one output offset, which will then be overridden by output_zero_point. We should probably remove one or the other to avoid confusion.

felix-johnny: Thanks!.. so, the regenerated data was unchanged..=> output_zero_point takes precedence..I'll push a new patch..

mansnils (approved):

Issue #1066, opened 2020-12-10
Mention 'Enable Trace' in Armv8.1's PMU example codes


tsunghanlin:
In the following example code sequence:
https://arm-software.github.io/CMSIS_5/Core/html/group__pmu8__functions.html

it doesn't mention we have to enable trace before using PMU.
However that is necessary and mentioned in the following user guide:
https://developer.arm.com/documentation/arm051-799564642-251/latest

PR #1081, opened 2020-12-31
DoxyGen: remind users to enable trace before using PMU

tsunghanlin:
Add some notes to remind users to enable trace via:
	CoreDebug->DEMCR |= CoreDebug_DEMCR_TRCENA_Msk;
before using PMU.

This is originally mentioned in
	'Application Note
	Armv8.1-M Performance Monitoring User Guide'


tsunghanlin:
This is for fix #1066

PR #1090, opened 2021-01-08
CMSIS-NN: Run unit test with FVP

mansnils:
* Add option to use FVP based Arm Corstone-300 software for unit testing.
* Update README with this information.


mansnils:
@felix-johnny could you review this?

felix-johnny:
Tried it out and looks good! Is there a way to run all the unit tests like it can be done using mbed-os?

mansnils:
@felix-johnny There is no official way but could do something like this in bash
```
FILES=TestCases/*/*.elf
FVP=Corstone_SSE-300_Ethos-U55
for f in $FILES
do
  echo "Testing $f.."
  $FVP $f
done
```

Issue #1092, opened 2021-01-13
CMSIS-NN: Potential data corruption for Avg pool - MVE only


mansnils:
In function compare_and_replace_if_larger_q7() it should be vctp8q() instead of vctp16q().

PR #1093, opened 2021-01-13
CMSIS-NN: Fix arm_max_pool_s8 for MVEI

mansnils:
* Replace vctp16q with vctp8q intrinsic.
* Update unit test as well to have const input data.

This is fixing: https://github.com/ARM-software/CMSIS_5/issues/1092


mansnils:
@felix-johnny could you review this?

PR #1099, opened 2021-01-20
CMSIS-NN: Update revision history

felix-johnny:
1. ARM.CMSIS.pdsc file updated with recent file changes
2. Updated revision history and version number
3. arm_nnfunctions.h file is clang formatted


JonatanAntoni (approved):
Looks good.

PR #1108, opened 2021-01-26
CMSIS-NN: Add support for null bias for Fully Connected op

felix-johnny:
NULL bias support is added for Fully Connected op
for  processors with and without DSP extension.

Added Test case to verify it

Also contribution by @mansnils


felix-johnny:
@mansnils Could you review this?

mansnils (changes requested):
batches was missing here. it should be batches before generate_bias.
..., batches, generate_bias) should be enough

CMSIS/NN/Tests/UnitTest/generate_test_data.py
```diff
@@ -669,9 +674,9 @@ class FullyConnectedSettings(TestSettings):
     def __init__(self, dataset, testtype, args, in_ch=1, out_ch=1, x_in=1, y_in=1, w_x=1, w_y=1, stride_x=1, stride_y=1,
                  pad=False, randmin=-7, randmax=7, outminrange=-128, outmaxrange=127, batches=1, input_scale=1.0,
                  input_zero_point=0, weights_scale=1.0, weights_zero_point=0, bias_scale=1.0, output_scale=1.0,
-                 output_zero_point=0):
+                 output_zero_point=0, generate_bias=True):
         super().__init__(dataset, testtype, args, in_ch, out_ch, x_in, y_in, w_x, w_y, stride_x, stride_y, pad, randmin,
-                         randmax, outminrange, outmaxrange, batches)
```
mansnils, line 674: This is the place I ment..
batches was missing here. it should be batches before generate_bias.
..., batches, generate_bias) should be enough

mansnils (approved):

PR #1120, opened 2021-02-04
CoreValidation: Updat build config and script.

JonatanAntoni:
- Add new build.py using python-matrix-builder module.
- Rename model configs to reflect build target.
- Add Cortex-M55 build targets.
- Clang/GCC use compiler features instead of -mfpu explicitly
- Remove disregarded flags


GuentherMartin (commented):
Looks fine.

Issue #1131, opened 2021-02-19
CMSIS-NN: Problem with clamping in max pool for MVEI


mansnils:
In CMSIS/NN/Source/PoolingFunctions/arm_max_pool_s8.c function clamp_output result is not clamped to min value.

PR #1132, opened 2021-02-19
CMSIS-NN: Fix bug in max pool for MVEI

mansnils:
This is a fix for: https://github.com/ARM-software/CMSIS_5/issues/1131


mansnils:
@felix-johnny Can you review this?

felix-johnny:
@mansnils could you add a TC as well for this?

mansnils:
@felix-johnny Ok will do

CMSIS/NN/Tests/UnitTest/TestCases/TestData/maxpooling_7/config_data.h
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+#define MAXPOOLING_7_OUT_CH 1
+#define MAXPOOLING_7_IN_CH 1
+#define MAXPOOLING_7_INPUT_W 4
+#define MAXPOOLING_7_INPUT_H 2
+#define MAXPOOLING_7_DST_SIZE 2
+#define MAXPOOLING_7_INPUT_SIZE 8
+#define MAXPOOLING_7_INPUT_OFFSET -128
+#define MAXPOOLING_7_OUTPUT_OFFSET 0
+#define MAXPOOLING_7_OUT_ACTIVATION_MIN 0
```
felix-johnny, line 29: Looking at the activation min/max all other test cases have the int8 min/max values.. And its good that this one has 0 as min.. but, would be good to have non-127 as max as well..could you update that either in this TC or in another one..?
 Also, could you make the output offset non-zero? all have zero right now.

mansnils: Ok, removing offsets for the pooling tests as agreed.

PR #1153, opened 2021-03-19
CMSIS-NN: Update FC test cases to use zero weight offset

felix-johnny:
The version history of packages are updated mainly
related to the use of tensorflow 2.4.0

Added a fixed download name for download of unity
tar ball.


felix-johnny:
@mansnils  Could you review this PR?

mansnils (commented):
We should probably also remove the option to set weights_zero_point and hard code self.weights_zero_point to 0 in FullyConnectedSettings

felix-johnny:
That is a good point.. Will fix that!

mansnils (approved):

PR #1159, opened 2021-03-24
CMSIS-NN: Refactor CMake

mansnils:
* Builds CMSIS-NN source the same way as when building unit tests and
  when building it standalone.
* Builds SVDF as well.


mansnils:
@felix-johnny can you review this?

felix-johnny (commented):
@christophe0606 Could you have a look at this change?

christophe0606 (approved):

Issue #1151, opened 2021-03-19
CMSIS-NN : Update weight offset to zero for Fully Connected


felix-johnny:
Currently non-zero weight offsets are supported for Fully Connected operator in CMSIS-NN but the quantization spec for int8 describes this as zero. 

DoD: Update FC operator for across all Arm Cortex-M processors to treat weight offset as zero.

PR #1161, opened 2021-03-25
CMSIS-NN : Update weight offset as zero in FC

felix-johnny:
Fully Connected implementation is updated to reflect
zero weight offset as per int8 spec.

MVE version and non-DSP extension versions are updated


felix-johnny:
@mansnils could you review the changes?

mansnils (approved):

felix-johnny:
Addresses part of https://github.com/ARM-software/CMSIS_5/issues/1151

Issue #294, opened 2017-12-14
Trigonometric functions in CMSIS-DSP


escherstair:
CMSIS-DSP offers only few trigonometric functions (`arm_sin`, `arm_cos` and `arm_sin_cos`, as far as I can see).
Some other functions are useful for my needs; I mean:

- [tangent](https://en.wikipedia.org/wiki/Trigonometric_functions#tangent) - I know I can calculate it using `arm_sin` and `arm_cos`

- [atan2](https://en.wikipedia.org/wiki/Atan2)

- `arcsin `and `arccos`

For the end developer it's quite difficult to implement the inverse trigonometric functions.
I'm almost interested in q31_t format.

Do you think they would be available in a future release?

llefaucheur:
Hi, among the various implementation options we can propose an algorithm optimised for speed using one of the series of the "Handbook of Mathematical Functions" (Abramowitz & Stegun).

escherstair:
Hi,
do you mean CMSIS-DSP library will be extended with this kind of functions?
Or are you suggesting me to implement them following the "Handbook of Mathematical Functions"?

llefaucheur:
Hi, We will implement the circular and inverse circular functions using formulas of this book. Regards.

pliniofpa:
Laurent,
With the answer above you are saying that CMSIS doesn't have plans to extende the math library to include atan, asin, acos, atan2, etc?

llefaucheur:
Hi pliniofpa, we will extend CMSIS_DSP in several directions, including mathematical functions. My point with escherstair, is to let you propose an algorithm based on public litterature.

DanSzaboStoneAge:
FWIW, I just realized that atan2 wasn't supported (oops), and ended up using the stdlib version.  I found [this article](https://www.dsprelated.com/showarticle/1052.php) which has a nifty approximation of atan2 with some further refinements in the comments, all of which looks pretty speedy.  It'd be interesting to compare with the LUT based atan implementation that seems to in progress.  Cheers.

christophe0606:
@DanSzaboStoneAge Thanks for feedback. Yes, Remez is a good method for finding approximations.
LUT can be ok too if the table is not too big.

DanSzaboStoneAge:
[...]

PR #1164, opened 2021-03-26
dsp: add atan function

shlomow:
Add atan fast implementaion function for
floating point representation.

This implements part of #294

### DIFF
--- a/CMSIS/DSP/Include/arm_common_tables.h
+++ b/CMSIS/DSP/Include/arm_common_tables.h
@@ -497,6 +497,11 @@ extern "C"
     extern const q15_t sinTable_q15[FAST_MATH_TABLE_SIZE + 1];
   #endif /* !defined(ARM_DSP_CONFIG_TABLES) defined(ARM_ALL_FAST_TABLES) */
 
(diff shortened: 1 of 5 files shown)


JonatanAntoni:

christophe0606:
@shlomow Thank you for the clean and detailed pull request.
I cannot currently merge it. We will need to spend more time on it.

There are lot of tradeoff to consider for functions like this one:
Do we want a look-up table ? Or an implementation based upon polynomial interpolation ?
What the accuracy that we want ? What's the tradeoff speed / accuracy which is needed ? Is it easy to vectorize the code ? What the acceptable size for the look-up table ?

So, before we start with such an implementation we will need to answer the above questions.

christophe0606:
I have just pushed an atan2 function (also available in Python wrapper from [pypi](https://pypi.org/project/cmsisdsp/1.4.0/)

It is using a polynomial interpolation because I did not want to use another big table (there are too many tables in the library).

But it means it is using more cycles than a simpler LUT based implementation.

For some application a faster (but perhaps less accurate) atan2 may be needed. So, it is just a first release. I may add other variants (like a faster one).

PR #1175, opened 2021-04-19
CMSIS-Core(M): Fix cache maintenance functions.

JonatanAntoni:
The addr argument needs to accept volatile void* in order
to be generic. In fact it doesn't affect the function
behavior which actual data type is used as only the
address of the data is required.


GuentherMartin (approved):
Looks correct to me.

PR #1184, opened 2021-04-28
CoreValidation: Fix Arm Compiler 6 linker warnings.

JonatanAntoni:
The compiler issues warning about unsupported (deprecated)
target architectures when called with specific -march.


GuentherMartin (approved):
Look good to me.

PR #1197, opened 2021-05-05
DSP: Removing deprecated library variant.

JonatanAntoni:


christophe0606 (approved):
From CMSIS-DSP point of view, it looks ok.

GuentherMartin (approved):
Looks fine to me.
Do we need to check if any DSP example is using the libraries?

JonatanAntoni:
Just checked by grep'ing over all uvprojx files: All the MDK examples are using the source variant, already.

PR #1203, opened 2021-05-27
CMSIS-Core(M): added instruction barrier (ISB) to set_CONTROL functions.

GuentherMartin:
PR solving request from Edmund Player raised in dsg-cmsis.


JonatanAntoni:
@EdmundPlayer, can you please cross check this.

JonatanAntoni:
@TTornblom, may I ask you to check if a similar change is required for IAR compiler?

EdmundPlayer:
> @EdmundPlayer, can you please cross check this.

Looks good to me :)
In theory, the ISB is not needed as much for the function that is intended to be executed from the secure world that changes the non-secure CONTROL_NS register, __TZ_set_CONTROL_NS(), but I think it doesn't hurt anyway to be on the safe side.

PR #1205, opened 2021-06-02
CMSIS-Core(M): added instruction barrier (ISB) to set_CONTROL functio…

TTornblom:
…ns (IAR).

Similar changes as PR 1203 but for IAR.


JonatanAntoni:

PR #1212, opened 2021-06-14
Device: Cleanup IRQ names to follow known standard.

JonatanAntoni:


vovamarch (changes requested):
Could you also align documentation to this naming, so that 'SVC' is used instead of 'SV Call' or 'SVCall'. See that Ref_NVIC.txt needs such chagnes and potentially also Template.txt

vovamarch (approved):

PR #1219, opened 2021-06-23
gen_pack: Add powershell version

paperchalice:
Powershell exists in most modern Windows computers, so users do not need to install extra tools like bash version on Windows.

This file may not be appropriate to put in the `Bash` folder.


grasci-arm:

JonatanAntoni:

JonatanAntoni:
Hi @paperchalice, 

Thanks for your contribution. 
You're right, the folder name "Bash" is not accurate once adding different generator scripts. We might rename the folder somewhen. But for the time being, I'd just put it here.

Do you volunteer to maintain this Powershell script? You may want to place your authors signature into its file header.

Cheers,
Jonatan

paperchalice:
Add document and my signature for this script.

JonatanAntoni (approved):

Issue #1226, opened 2021-06-30
DSP: Add float 64 bits to basic functions


jeanbie0:
Hello,

Some basics functions like Vector Substraction (https://arm-software.github.io/CMSIS_5/DSP/html/group__BasicSub.html) (arm_sub_fxx) are in float 16 and 32 but not in float 64 bits.

I'm doing some of these functions because I had to use it with a compatible double floating point unit processor. Do you know I can push these new files? It'll be more easy to organize my jobs and others people will not have to redo the job.

Thank you

JonatanAntoni:
Hi @jeanbie0,

Thanks for getting in touch. Basically, we are open for receiving contribution via Pull-Requests.
I added in @christophe0606 who maintains the DSP part of CMSIS.

Cheers,
Jonatan

christophe0606:
Hi @jeanbie0,

Indeed we are missing some f64 code.

The problem is that for f16, f32 and f64 code, we have lot of code duplication for the scalar code. It is basically the same code but with different types and constants.

I am thinking about ways to avoid this. 

But if you don't have a lot of new f64 functions, then I can perhaps merge them even if it means more code duplication.

christophe0606:
Hi @jeanbie0,

We have recently added more f64 functions. We are not covering 100% of the existing f32 functions so obviously more will be added in the future.

But we are covering lots of important functions already.

So, I am closing this ticket.

PR #1227, opened 2021-07-01
Add some f64 DSP functions

jeanbie0:
Hello,

This pull request is related to the issue #1226.

It add some float64 functions (the ones I needed):

- All BasicMathFunctions
- arm_sqrt_f64
- arm_cmplx_mag_f64

Like you explain in the issue, it's mostly the same code with type modification.

Regards

Jean

### DIFF
--- a/CMSIS/DSP/Include/dsp/basic_math_functions.h
+++ b/CMSIS/DSP/Include/dsp/basic_math_functions.h
@@ -2,7 +2,7 @@
  * @file     basic_math_functions.h
  * @brief    Public header file for CMSIS DSP Library
  * @version  V1.9.0
- * @date     23 April 2021
+ * @date     1 July 2021
  * Target Processor: Cortex-M and Cortex-A cores
  ******************************************************************************/
 /*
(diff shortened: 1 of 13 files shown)


grasci-arm:

christophe0606:
Hi @jeanbie0,

Thanks for this pull request.

I have looked at the code. 

There are a few remaining issues:
1 - You converted the vector code but we are not targeting any core with f64 vectors in CMSIS-DSP so it will not work. So, you need to remove all of the vector code from those files and keep only the scalar code ;

2 - arm_sqrt_f64 has some issues.
It is using f64 asm not supported on the cores we target.
There are still use of sqrtf and x.yyf constants.

Note that if I merge this pull request it will not be in the coming weeks because I am working on other topics currently. And, I'll also have to write some tests before I can merge those functions.

So, you'll have to be patient. 

But, if you have to wait and I plan on refactoring f16, f32 and f64 code to avoid duplication, can you wait a bit more and use my refactoring with the new f64 functions ? In that case, I would not merge this pull request.

Thanks.

jeanbie0:
Hi @christophe0606 ,

Thank for your review. I only use scalar code so I copy/paste vector code.
I'll wait for your refactoring it'll be easier and more efficient.

christophe0606:
Issue #1226 was closed so I close this pull request too.

More f64 functions are now available and more will be added in the future but we are already covering lots of important functions.

PR #1231, opened 2021-07-01
Update CMSIS-Build Documentation

brondani:
Remove deprecated makefile generation and ccmerge.


jkrech (commented):
LGTM. I was wondering whether we should also consider to remove cbuild_uv due to it's dependency on MDK installation.

jkrech (approved):
LGTM. Removing this script from the open source CMSIS-Build makes it vendor independent.

PR #1238, opened 2021-07-09
rtx-def: fix license - apache 2.0

0xc0170:
Using the same license as any other file in RTX. 

I found this while working with 5.8.0 release (our license check in Mbed OS failed due to non permissive license found). It would be good to have license check introduced to gate PRs.


grasci-arm:

RobertRostohar:
Thanks for pointing this out and providing a pull request.

PR #1250, opened 2021-07-16
Fix PMU_Type spurious reserved register and offset documentation.

ithinuel:
This fixes inconsistencies between the documentation available [here](https://developer.arm.com/documentation/101051/0100/Performance-Monitoring-Unit-Extension/PMU-register-summary) and the definition of `PMU_Type`.


grasci-arm:

PR #1251, opened 2021-07-17
people forgot to update the sub version...

GorgonMeducer:
The CMSIS core version should be 5.5.0, and people forgot to update the cmsis_version.h
We should avoid this issue in the future.


grasci-arm:

JonatanAntoni:
Hi @GorgonMeducer,

Thanks for pointing this out. You are apperently right and this totally slipped through. Damn.

-Jonatan

GorgonMeducer:
I have seen significant differences between CMSIS5.7.0 and CMSIS 5.8.0. The only way to write a compatible code is using the sub version macro...
The moment I notice this bug, how desperate I was...Orz.

Thank you for accepting this merge request.

Cheers

PR #1253, opened 2021-07-20
CMSIS-NN: Cleanup of CMSIS-DSP dependencies

felix-johnny:
Left over CMSIS-DSP dependencies are removed

For functions with just DSP extension optimization,
the reference C implementation(if available) is picked
for processors with Helium Technology as auto vectorization
can potentially give a better performance. This affects only
the non-TFLM functions.

### DIFF
--- a/CMSIS/NN/Include/arm_nnsupportfunctions.h
+++ b/CMSIS/NN/Include/arm_nnsupportfunctions.h
@@ -21,8 +21,8 @@
  * Title:        arm_nnsupportfunctions.h
  * Description:  Public header file of support functions for CMSIS NN Library
  *
- * $Date:        5. July 2021
- * $Revision:    V.5.6.0
+ * $Date:        20. July 2021
+ * $Revision:    V.5.7.0
  *
  * Target Processor:  Cortex-M CPUs
  * -------------------------------------------------------------------- */
@@ -47,6 +47,13 @@ extern "C" {
 #define MIN(A, B) ((A) < (B) ? (A) : (B))
 #define CLAMP(x, h, l) MAX(MIN((x), (h)), (l))
 #define REDUCE_MULTIPLIER(_mult) ((_mult < 0x7FFF0000) ? ((_mult + (1 << 15)) >> 16) : 0x7FFF)
+
+/**
+ * @brief definition to pack four 8 bit values.
+ */
+#define PACK_Q7x4_32x1(v0, v1, v2, v3)                                                                                 \
+    ((((int32_t)(v0) << 0) & (int32_t)0x000000FF) | (((int32_t)(v1) << 8) & (int32_t)0x0000FF00) |                     \
+     (((int32_t)(v2) << 16) & (int32_t)0x00FF0000) | (((int32_t)(v3) << 24) & (int32_t)0xFF000000))
 
 /**
  * @brief Union for SIMD access of q31/q15/q7 types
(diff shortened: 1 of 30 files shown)


grasci-arm:

CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -539,6 +546,18 @@ __STATIC_FORCEINLINE q31_t arm_nn_read_q7x4(const q7_t *in_q7)
     return (val);
 }
 
+/**
+  @brief         Write four q7 to q7 pointer and increment pointer afterwards.
+  @param[in]     in       Double pointer to input value
+  @param[in]     value    Four bytes to copy
+  @return        none
+ */
+__STATIC_FORCEINLINE void arm_nn_write_q7x4_ia(q7_t **in, q31_t value)
```
christophe0606, line 555: It is very similar to CMSIS-DSP write_q7x4_ia. At some point we should try to identify the common functions we have and export them perhaps to CMSIS-Core.

The only difference is that CMSIS-DSP is still having a #if __ARM_FEATURE_UNALIGNED for the memcpy although memcpy can handle unaligned accesses. But I had observed crashes on M0 with some compilers. I'll need to spend some time on this to understand the root cause and remove this #if But CMSIS-DSP still has to support some older compilers ...

CMSIS/NN/Source/ActivationFunctions/arm_relu_q7.c
```diff
@@ -66,15 +66,15 @@ void arm_relu_q7(q7_t *data, uint16_t size)
 
     while (i)
     {
-        in = read_q7x4_ia(&input);
+        in = arm_nn_read_q7x4_ia((const q7_t **)&input);
```
christophe0606, line 69: Why is input not defined as const ?

CMSIS/NN/Source/BasicMathFunctions/arm_elementwise_add_s8.c
```diff
@@ -209,7 +206,7 @@ arm_status arm_elementwise_add_s8(const int8_t *input_1_vect,
         sum = MIN(sum, out_activation_max);
         r4 = (q7_t)sum;
 
-        write_q7x4_ia(&output, __PACKq7(r1, r2, r3, r4));
+        arm_nn_write_q7x4_ia(&output, PACK_Q7x4_32x1(r1, r2, r3, r4));
```
christophe0606, line 209: Another common feature NN and DSP both need. DSP has __PACKq7
Another candidate for moving it somewhere else ?

christophe0606 (approved):

CMSIS/NN/Source/ActivationFunctions/arm_relu_q7.c
felix-johnny: It is one of the pre-TFLM API's where the output location is same as the input.

Most operators now have the activation function fused and this is not used.

CMSIS/NN/Include/arm_nnsupportfunctions.h
felix-johnny: Yes, there is scope to combine things here for the bunch of read/write_q7x4/q15x2_ia/no_ia API's . There is a const difference in the input of most when compared to the ones in DSP..

CMSIS/NN/Source/BasicMathFunctions/arm_elementwise_add_s8.c
felix-johnny: Yes.. this too along with the read/write ones..

PR #1255, opened 2021-07-21
Update core_cm33.h

AlbertHuang-CPU:
Add SFSR and SFAR in SCB_Type.


grasci-arm:

JonatanAntoni:
Hi @AlbertHuang-CPU,

We checked the header for these registers and it looks like they are already part of the `SAU_Type`. This is aligned with the Arm®v8-M Architecture Reference Manual.

Is it a fair statement that is should be enough to have the registers defined there? Or are there any specific requirements to have the registers being part of the `SCB_Type` in addition?

Thanks,
Jonatan

AlbertHuang-CPU:
Hi Jonatan,

In Arm v8m definition, the address is just in the SCB scope.
***@***.***D77E62.79588C60]
In practical, SAU can be configured out, and be taken over by implementation defined IDAU.
That means although there is no SAU, the secure fault still possible happens and the SFSR still need to be accessed.

Regards,
Albert

grasci-arm:

PR #1281, opened 2021-08-13
CMSIS-NN: Optimized convolution for 16x8 data type

oscarandersson8218:
Authors: Sebastian-Larsson, oscarandersson8218 (GitHub)

1. Optimized function is added for input data type of int16
and filter data type of int8. The main constraint is
that the kernel volume is < 512.

2. Unit tests are added as well.

Change-Id: I9f0f3067a275c8bf5f9263ebe76cd9a1033c7793


grasci-arm:

felix-johnny (approved):

Issue #1047, opened 2020-11-20
NN_ROUND() issues in 16bit ops 


majianjia:
I met an issue of 16 bit NN producing bad results when I tried to use `round()` in NN calculation using `ARM_NN_TRUNCATE` (undefined ARM_NN_TRUNCATE)
I normally define ARM_NN_TRUNCATE but recently I want to try to use round so I undefined the macro. In 8 bit ops, it brings some accuracy improvement but in 16 bit it fails. 

The 16bit ops I use is this one but other which uses NN_ROUND() have similar issues. 
https://github.com/ARM-software/CMSIS_5/blob/e42a76a56841fed4dbd182221cc7f3ade753acc9/CMSIS/NN/Source/FullyConnectedFunctions/arm_fully_connected_mat_q7_vec_q15_opt.c#L117

change `#define NN_ROUND(out_shift) ( (0x1u << out_shift) >> 1 )` to 
`#define NN_ROUND(out_shift) ((0x1 << out_shift) >> 1 )` fixed my issue. 

https://github.com/ARM-software/CMSIS_5/blob/e42a76a56841fed4dbd182221cc7f3ade753acc9/CMSIS/NN/Include/arm_nnsupportfunctions.h#L652

Also, it doesn't seem right when it adds to a negative number. Should the round be round down in negative number? In most ops, only add this to the bias which means it is always positive. 

More information will be provided when I figure it out. Thanks.

felix-johnny:
@majianjia Apologies for the delay. Would you mind making a PR for the fix?

majianjia:
@felix-johnny let me reminding myself what is this about.
Happy to, thanks :p

felix-johnny:
Great , thanks!

majianjia:
close issue as the PR is merged. 
Thanks

PR #1295, opened 2021-08-25
CMSIS-NN: Fixed bug in NN_ROUND()

majianjia:
Fixed bug related to ARM_NN_TRUNCATE does not perform correctly with 16-bit ops. 
https://github.com/ARM-software/CMSIS_5/issues/1047


felix-johnny:
@majianjia Thanks for the PR!

PR #1299, opened 2021-09-07
DoxyGen: Fixed documentation of cache maintenance functions.

JonatanAntoni:
In c0e1824 the data type of the addr parameters were changed
to volatile void* but the documentation got missed.


KeilChris (approved):
Looks good to me!

PR #1301, opened 2021-09-07
DAP: refactor clock delay calculation

flit:
The clock delay calculation code is refactored into `Set_DAP_Clock_Delay()`, which is called from both `DAP_Setup()` and
`DAP_SWJ_Clock()`.

This change allows the default clock frequency to use fast clock mode. It was required for the micro:bit v2 firmware.

cc @gerargz @mbrossard @microbit-carlos


grasci-arm:

PR #1306, opened 2021-09-09
CMSIS-DSP: Correct interpol(...) arg in ComplexMathFunctions CMakeLists.txt

sandeepmistry:
Without these changes I would get the following error when compiling my CMake project with `cmake .. -DCONFIGTABLE=ON -DRFFT_Q15_256=ON -DARM_COS_F32=ON`:

```
CMake Error at lib/CMSIS_5/CMSIS/DSP/Source/interpol.cmake:4 (target_compile_definitions):
  Cannot specify compile definitions for target "CMSISDSPFastMath" which is
  not built by this project.
Call Stack (most recent call first):
  lib/CMSIS_5/CMSIS/DSP/Source/ComplexMathFunctions/CMakeLists.txt:15 (interpol)
```

I believe `CMSIS/DSP/Source/ComplexMathFunctions/CMakeLists.txt` is using `interpol(CMSISDSPFastMath)` by mistake when it should be using `interpol(CMSISDSPComplexMath)` instead.

@christophe0606 what do you think?

### DIFF
--- a/CMSIS/DSP/Source/ComplexMathFunctions/CMakeLists.txt
+++ b/CMSIS/DSP/Source/ComplexMathFunctions/CMakeLists.txt
@@ -12,7 +12,7 @@ configDsp(CMSISDSPComplexMath ${ROOT})
 
 
 include(interpol)
-interpol(CMSISDSPFastMath)
+interpol(CMSISDSPComplexMath)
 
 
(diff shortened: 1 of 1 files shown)


grasci-arm:

christophe0606:
@sandeepmistry  You're right. This is a bug. Thank for finding it.

PR #1307, opened 2021-09-13
Implement Float64 support for some DSP functions (draft)

jbr-smtg:
Hello all,

The following pull request will probably be refused since it requires a lot of modifications.
However this is the first step toward adding more float64 basic support for some DSP functions.
Don't hesitate to let me know everything that needs to be changed/modified.

Thanks in advance for your time!

PR #1308, opened 2021-09-13
Implement Float64 support for some DSP functions 

jbr-smtg:
Hello all,

The following pull request is similar to https://github.com/ARM-software/CMSIS_5/pull/1307 but contains the following fixes:
- Branch Rebased
- Fix new files version number

Don't hesitate to let me know everything that needs to be changed/modified.

Thanks in advance for your time!


grasci-arm:

CMSIS/DSP/Source/BasicMathFunctions/arm_dot_prod_f64.c
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+void arm_dot_prod_f64(
+  const float64_t * pSrcA,
+  const float64_t * pSrcB,
+        uint32_t blockSize,
+        float64_t * result)
+{
+  uint32_t blkCnt;                               /* Loop counter */
+  float64_t sum = 0.0f;                          /* Temporary return variable */
```
christophe0606, line 69: The f suffix must be removed. All constants must be double in f64 code.
There are probably a lot of other files with similar problems.

CMSIS/DSP/Source/FilteringFunctions/arm_fir_f64.c
```diff
@@ -0,0 +1,231 @@
...
+ * limitations under the License.
+ */
+
+#include "dsp/filtering_functions.h"
+
+/**
+  @ingroup groupFilters
+ */
+
+/**
```
christophe0606, line 35: This should be removed. The doxygen group is defined already by the f32 function.
I think there may be similar problems with other functions.

Generally the doxygen group is introduced by the f32 function and should not be duplicated in the f64.

christophe0606 (changes requested):
In the previous (closed) pull request there were some cmake updates.
I no more see them.

The new files should be added to the cmake if the cmake is not using a file(GLOB SRCF64 "./*_f64.c")

CMSIS/DSP/Source/BasicMathFunctions/arm_dot_prod_f64.c
jbr-smtg: Thanks! I just fixed it.

CMSIS/DSP/Source/FilteringFunctions/arm_fir_f64.c
jbr-smtg: I removed the docygen group.

christophe0606:
@jbr-smtg But you have the problem of float constant and doxygen group in much more than the files I have highlighted.

For the doxygen defgroup I think you may have the problem in nearly all the f64 files if they come from a copy of the f32 file.

For the float issue, each f32 file containing a constant will give the issue.

For instance, arm_mean_f64.c has the issue (0.0f constant to init the sum).

I have not reviewed all the files for those problems.

CMSIS/DSP/Source/BasicMathFunctions/arm_dot_prod_f64.c
jbr-smtg: I did a new commit that should have fixed most of them.

christophe0606:
@jbr-smtg Can you add the new files to the CMakeLists.txt (except when the CMakeLists.txt are using a pattern like file(GLOB SRCF64 "./*_f64.c") in BasicMaths).

Because I need to try to rebuild with a few different compilers and targeting several different cores before I can accept the pull request. On older compilers we often have some surprises requiring small changes to the code for building.

I'll accept the pull request only when I can check it is building on all our configurations and that the documentation generated from the Doxygen is correct.

Thank you.

jbr-smtg:
Thank-you :)

christophe0606:
@jbr-smtg I have found some additional issues:

**arm_max_no_idx_f64**

`float64_t   maxValue = F32_MIN`

It should be F64_MIN.

**arm_chebyshev_distance_f64**

There is a remaining 0.0f

After those changes, I should be able to merge the pull request and start working on the tests.

christophe0606:
Merged.

Now I am going to work on the tests.

jbr-smtg:
Perfect :)
Thanks again!

PR #1317, opened 2021-09-27
DSP: Eliminate some non-standard C code

TTornblom:
Cleaned out some non-standard C code, typeof() and using
"const" variables as constants.

Also added some initial IAR support.


grasci-arm:

PR #1322, opened 2021-09-30
Adding Support for Apple Compiler

jbr-smtg:
Hello all,

The Apple XCode compiler uses the macro __APPLE_CC__ to define its presence.
I would like to suggest defining the macros for this compiler too in order to provide basic support.

Thanks in advance!

### DIFF
--- a/CMSIS/DSP/Include/arm_math_types.h
+++ b/CMSIS/DSP/Include/arm_math_types.h
@@ -98,6 +98,10 @@ extern "C"
 
 #if !defined(ARM_MATH_AUTOVECTORIZE)
 
+#if !defined(__ARM_FEATURE_MVE)
+#define __ARM_FEATURE_MVE 0
+#endif
+
 #if __ARM_FEATURE_MVE
   #if !defined(ARM_MATH_MVEI)
     #define ARM_MATH_MVEI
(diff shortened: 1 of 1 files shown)


grasci-arm:

CMSIS/DSP/Include/arm_math_types.h
```diff
@@ -161,6 +165,12 @@ extern "C"
   #define LOW_OPTIMIZATION_EXIT
   #define IAR_ONLY_LOW_OPTIMIZATION_ENTER
   #define IAR_ONLY_LOW_OPTIMIZATION_EXIT
+  
+#elif defined ( __APPLE_CC__ )
```
christophe0606, line 169: Is \_\_GNUC\_\_ implied by \_\_APPLE\_CC\_\_ ?
Because at beginning and end of the file there are lots of compiler tests and there is no tests for \_\_APPLE\_CC\_\_.

The tests are beginning of file are ending with #error Unknown compiler
So if you can build it means the tests are passing and one of the symbol (\_\_GNUC\_\_ ?) must be defined ?

CMSIS/DSP/Include/arm_math_types.h
```diff
@@ -98,6 +98,10 @@ extern "C"
 
 #if !defined(ARM_MATH_AUTOVECTORIZE)
 
+#if !defined(__ARM_FEATURE_MVE)
```
christophe0606, line 101: Why is it needed here ?
If you build with Neon enabled, you have a test:

`#if __ARM_FEATURE_FP16_VECTOR_ARITHMETIC`

So you may have a similar problem ?

Perhaps it is better to use

`#if defined(__ARM_FEATURE_MVE)` 

before the tests for the \_\_ARM_FEATURE_MVE value.

CMSIS/DSP/Include/arm_math_types.h
jbr-smtg: Indeed, __GNUC__ is implied by __APPLE_CC__
However the attribute __attribute__(( optimize("-O1") )) used in case of __GNUC__  is a little problematic with Xcode:
"Unknown attribute 'optimize'"

christophe0606:
I have made some small modifications to this pull request and pushed the changes in the new commit https://github.com/ARM-software/CMSIS_5/commit/ad72032487fb55313e06cecfb4a1b50a83cde17d

So I close this pull request. Please check that my latest commit is solving the issue and is equivalent to this pull request.

PR #1325, opened 2021-09-30
Fixing warnings issues with Apple Compiler/Xcode

jbr-smtg:
Hello all,

There are a couples of occurrence of a warnings with the Apple Compiler. The IDE Xcode suggests the fix in this pull request.

Thanks in advance!

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -1,8 +1,8 @@
 /**************************************************************************//**
  * @file     cmsis_gcc.h
  * @brief    CMSIS compiler GCC header file
- * @version  V5.4.1
- * @date     27. May 2021
+ * @version  V5.4.2
+ * @date     30. September 2021
  ******************************************************************************/
 /*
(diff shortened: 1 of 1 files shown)


jbr-smtg:
I can see plenty of compilation tests failing with the following error:
../../Core/Include/cmsis_gcc.h:41:50: error: invalid 'asm': invalid operand for code 'w'
Does that mean that this fix will not have a chance due to the "w" not being supported by the used compilers?

JonatanAntoni:
> Does that mean that this fix will not have a chance due to the "w" not being supported by the used compilers?

Yes, GCC seems not to support this. It looks like we'd need to add this as a separate compiler header only used for (Apple) Clang.

christophe0606:
@jbr-smtg Sorry for the naïve question here but : why is the file cmsis_gcc.h in Core modified and not in Core_A since you are trying to build for Cortex-A.

You should only need to modify the Core_A/Include/cmsis_gcc.h
The other file in Core does not need to be modified.

So in your XCode build you probably need to include Core_A instead of Core.

jbr-smtg:
I created a new pull request using a new file

PR #1327, opened 2021-10-01
DSP: Cleanup of type mismatches

TTornblom:
The IAR compiler appears to provide more warnings about type
mismatches than the other compilers. This cleans a lot of them up.

Also cleaned out some trailing white space in the updated files.


grasci-arm:

PR #1334, opened 2021-10-05
Ignore old c cast style not supported by apple cc

jbr-smtg:
Hello,

One more step to support Apple C Compiler: we need to add a specific ignore warning related to old c cast style.

Thanks again!


grasci-arm:

PR #1335, opened 2021-10-05
Fix warning unused tmp variable

jbr-smtg:
Hello all,

Please find here a fix for a warning happening in arm64 where a tmp variable is defined but not used.

Thanks again!


grasci-arm:

PR #1338, opened 2021-10-06
CMSIS-NN: Update softmax, svdf and pooling unit tests

mansnils:
Data generate script uses TFL int kernels instead of TF float kernels.
Softmax and pooling test cases use Keras for generating a model.
Svdf use a json template for generating a model.
Corresponding input headers are regenerated.


grasci-arm:

PR #1339, opened 2021-10-07
CMSIS-NN: Disable loop unroll in DW Conv

felix-johnny:
Loop unroll of already unrolled loop in
depthwise_conv_s8_mult_4 is disabled for clang


grasci-arm:

PR #1348, opened 2021-10-19
Fix read-out-of-bounds bug in SVDF kernel

jenselofsson:
The size argument to memmove reads one element out of bounds. This PR fixes that bug.


grasci-arm:

felix-johnny (approved):

PR #1357, opened 2021-11-01
CMSIS-NN: CMake updates

mansnils:
Sources added to the cmsis-nn static library should with added with
scope PRIVATE. Else they will be exported and built by the other
target linking the library.

Compile options should be set per target.

Change-Id: Ib5c5ce1db8f7baa28bba411be7eea955f5a3af29


grasci-arm:

felix-johnny (approved):

PR #1360, opened 2021-11-05
CMSIS-NN: Add Null bias check for DW conv

felix-johnny:
Missing null bias check is added to DW conv along
with unit tests

Change-Id: Ib2f817fd96c4edbde4667fa4cb3419604d118a03


grasci-arm:

mansnils (approved):

PR #1362, opened 2021-11-08
CMSIS-NN: Add restrict keyword

mansnils:
Add restrict for parameters of arm_nn_mat_mult_kernel_s8_s16 for
Armclang only.

### DIFF
--- a/CMSIS/NN/Include/arm_nnsupportfunctions.h
+++ b/CMSIS/NN/Include/arm_nnsupportfunctions.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2010-2021 Arm Limited or its affiliates. All rights reserved.
+ * Copyright (C) 2010-2021 Arm Limited or its affiliates.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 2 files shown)


grasci-arm:

CMSIS/NN/Include/arm_nnsupportfunctions.h
```diff
@@ -54,6 +54,12 @@ extern "C" {
     ((((int32_t)(v0) << 0) & (int32_t)0x000000FF) | (((int32_t)(v1) << 8) & (int32_t)0x0000FF00) |                     \
      (((int32_t)(v2) << 16) & (int32_t)0x00FF0000) | (((int32_t)(v3) << 24) & (int32_t)0xFF000000))
 
+#if defined(__ARMCC_VERSION)
```
felix-johnny, line 57: __RESTRICT is defined in cmsis_gcc/armclang.h files.. Let's use that instead. And are we sure to use restrict only for armclang?

felix-johnny:
Not taking this in as it isn't a generic solution.

PR #1363, opened 2021-11-08
 CMSIS-DSP : faster Q.15/Q.31 Helium matrix multiplications. Uses an …

FabKlein:
…initial transpose stage, requiring extra scratch space to hold RHS transposed matrix.


grasci-arm:

PR #1364, opened 2021-11-11
CMSIS-NN: Update unit test data generate script

mansnils:
* Take out dimensions from TFL reference.
* Create new testdata set folder for new tests.
* Remove all params.txt data as it is not needed any more.
* Add dilation support when setting padding values.


grasci-arm:

PR #1367, opened 2021-11-15
DSP: Add IAR tool chain

TTornblom:
Previously known as PR #1337

### DIFF
--- a/CMSIS/DSP/Platforms/FVP/ARMCM0/Include/system_ARMCM0.h
+++ b/CMSIS/DSP/Platforms/FVP/ARMCM0/Include/system_ARMCM0.h
@@ -2,11 +2,11 @@
  * @file     system_ARMCM0.h
  * @brief    CMSIS Device System Header File for
  *           ARMCM0 Device
- * @version  V5.3.1
- * @date     09. July 2018
+ * @version  V5.3.2
+ * @date     15. November 2019
  ******************************************************************************/
 /*
- * Copyright (c) 2009-2018 Arm Limited. All rights reserved.
+ * Copyright (c) 2009-2019 Arm Limited. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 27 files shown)


grasci-arm:

christophe0606:
Does it mean it needs to be reviewed again and my latest comments were lost ? Or they have already been taken into account in this new PR ?

TTornblom:
@christophe0606 
Were there other comments than what can be seen in #1337?
As far as I know there were no outstanding issues in #1337.

christophe0606:
@TTornblom It is the same comments.

But from those comments and the code I have question : was it tested on a fast model ? Did you build the CMSIS-DSP test framework with IAR tool chain and you was able to run it on a fast model ?

From the comments I am expecting some issues that's why I had not merged the PR. so far.

TTornblom:
We have sucessfully built and run the examples on our M55 simulator using the FVP platform. We have not attempted to run the test suites.

christophe0606:
@TTornblom It is not a problem if there is no support for IAR toolchain in the test suite. The test suite does not have to support all compilers since it is mainly used internally to validate the algorithms.

So should I close this PR ? It will require much more work if you really want to support the test suite.

TTornblom:
I would like to have this PR merged, not closed without action.

It adds DSP support for the M55 with IAR. Without it compilation fails and with it the examples all builds and runs on the IAR M55 simulator, which also successfully runs the examples built with GNUARM, so it is fairly accurate.

christophe0606:
@TTornblom Ok. Since it is also modifying a few files which are not only for IAR, I'll have to re-check this part and then I'll merge it.

christophe0606:
@TTornblom Can you just answer the comments related to f16 ?
Because I have had some github issue opened related to f16 and IAR.

TTornblom:
@christophe0606 I think the f16 issues were solved by you previously, so this patch is just for allowing the examples to be built and run with the IAR toolchain.

PR #1369, opened 2021-11-16
Build: Missing M55 include path

TTornblom:
Fixes issue causing compilation errors on IAR

### DIFF
--- a/ARM.CMSIS.pdsc
+++ b/ARM.CMSIS.pdsc
@@ -2490,6 +2490,7 @@ and 8-bit Java bytecodes in Jazelle state.
       <files>
         <!-- include folder / device header file -->
         <file category="include"      name="Device/ARM/ARMCM55/Include/"/>
+        <file category="include"      name="Device/ARM/ARMCM55/Include/Template"/>
         <!-- startup / system file -->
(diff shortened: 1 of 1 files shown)


grasci-arm:

ARM.CMSIS.pdsc
```diff
@@ -2490,6 +2490,7 @@ and 8-bit Java bytecodes in Jazelle state.
       <files>
         <!-- include folder / device header file -->
         <file category="include"      name="Device/ARM/ARMCM55/Include/"/>
+        <file category="include"      name="Device/ARM/ARMCM55/Include/Template"/>
```
JonatanAntoni, line 2493: This should not be required and may even break projects.
The `partition_ARMCM55.h` file is declared as a config file below. Hence is needs to be copied into the user project to be modifiably.

JonatanAntoni (changes requested):
This change does not make sense. What does not work in IAR with the current component?

ARM.CMSIS.pdsc
TTornblom: Sorry Jonatan, it was a secure/non-secure mismatch in my project.
I'll close the PR.

PR #1379, opened 2021-12-14
CMSIS-NN: Use single rounding

mansnils:
Adapt to Tflite/TfliteMicro, which will use single rounding instead of double rounding:
https://github.com/tensorflow/tensorflow/pull/50290

Changes are added under flag CMSIS_NN_USE_SINGLE_ROUNDING, which is
disabled by default.


grasci-arm:

felix-johnny (approved):

Issue #1382, opened 2021-12-19
[BUG] `r1` overriden in SVC, GCC ignores it when compiling with LTO and Ofast


RisinT96:
Encountered the bug when trying to allocate two semaphores with the same arguments.
 
Inside the SVC multiple operations are performed:
* Finds the stack pointer of the calling thread
* Extracts the stacked registers {r0-r3,r12}
* Calls the service function
* Stores {r0, r1} into the stacked registers of the calling thread
 
Once the SVC returns, the stacked registers are loaded into the cpu.
 
Thus r0 and r1 have been overridden, in the case of allocating semaphores, r0 contains the semaphore address and r1 appears to contain 0.
 
When compiling with LTO, in some specific cases GCC seems to ignore the fact that r1 has changed and loads it back into r0 before the second SVC call, resulting in both arguments being 0 and the mutex allocation to fail.
 
I created the simplest example I could to demonstrate the bug, had to create my own SVC, as it happens only when the semaphore allocation is inlined by LTO, which is hard to enforce in a tiny project. 
I simply expanded the macros that make up `SVC0_3(SemaphoreNew,      osSemaphoreId_t, uint32_t, uint32_t, const osSemaphoreAttr_t *)` and removed the attributes argument as it seems to bear no effect.
 
Here is my example code that demonstrates the bug:
```c
#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>
 
static void *rtos_call_svc(uint32_t max_count, uint32_t initial_count)
{
    return (void *)0xDEADBEEF;
}
 
void *rtos_call(uint32_t a1, uint32_t a2)
{
    register uint32_t __r0 __asm("r0") = (uint32_t)a1;
    register uint32_t __r1 __asm("r1") = (uint32_t)a2;
    register uint32_t __rf __asm("r12") = (uint32_t)rtos_call_svc;
    /* Notice how `r1` is not mentioned in the `OutputOperands` nor the `Clobbers` */
    __asm volatile("svc 0"
                   : "=r"(__r0)
                   : "r"(__rf), "r"(__r0), "r"(__r1)
                   :);
    return (void *)__r0;
}
 
int main(void)
{
    // Doesn't work:
    // * r1 and r0 are overriden inside SVC
    // * r1 is loaded into r0
    // * the second call is executed with incorrect arguments
    void *res_1 = rtos_call(0x31337, 0x31337);
 
    if (res_1 == NULL)
    {
        for (;;)
        {
        }
[...]

PR #1383, opened 2021-12-19
Signalling GCC that both `r0` and `r1` are modified by `svc 0` call

RisinT96:
Fixed problems described in #1382.

I tried adding `r0` and `r1` to the `Clobbers` list when they weren't in the `OutputOperands` list, but it spewed out some errors if they appeared in the `InputOperands` list, I had to add them to the `OutputOperands` list.


grasci-arm:

RobertRostohar:
Thanks for this PR, however it does not fit since we plan to change the behavior of SVC to store only register "r0".

RisinT96:
> Thanks for this PR, however it does not fit since we plan to change the behavior of SVC to store only register "r0".

I can update my PR to do just that.

RobertRostohar:
Thanks for you help, however I suggest that we take this over. The handlers for Cortex-A are a bit more complex and we also need to update the version numbers, release notes, ...

PR #1403, opened 2022-01-20
DSP: Clean up "const **" vs "**" casts

TTornblom:
There are lots of casts between const and non-const pointers,
which may yield undefined behavior.

According to the strict aliasing rules for C the types
int * and const int * are not compatible, that implies that
updating an object of type const int * by dereferencing a
pointer of type int * is undefined.

This patch cleans a lot of this up.


grasci-arm:

TTornblom:
@christophe0606 Hello Christophe, can you please review this?

christophe0606:
@TTornblom It is generating too many compile errors with clang (and I have not yet tested with gcc).
Can you ensure that it is not breaking the clang and gcc build ?

TTornblom:
Need to go over this, closing this PR.

PR #1422, opened 2022-02-17
Update Jenkinsfile

JonatanAntoni:
- Explicit K8S agent for each slave
- Fetch Docker images from Artifactory
- Use --slice feature of build.py


JonatanAntoni:
Outdated

PR #1423, opened 2022-02-22
CMSIS-NN: Update documentation

felix-johnny:
Minor document update to explicitly mention
that it is for Cortex-M processors.

### DIFF
--- a/CMSIS/NN/Include/arm_nnfunctions.h
+++ b/CMSIS/NN/Include/arm_nnfunctions.h
@@ -21,8 +21,8 @@
  * Title:        arm_nnfunctions.h
  * Description:  Public header file for CMSIS NN Library
  *
- * $Date:        22 February 2022
- * $Revision:    V.8.1.0
+ * $Date:        10 March 2022
+ * $Revision:    V.8.1.1
  *
  * Target Processor:  Cortex-M CPUs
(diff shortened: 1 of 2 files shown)


mansnils (approved):

EdmundPlayer (commented):
Looks much better... I was wondering was 'without SIMD' means.  Do you just mean functions that don't include instructions provided by the DSP Extension?  

If so, maybe change the text to this?

Supported Processors
   * -------
   * CMSIS-NN supports different M-profile implementations. Typically, there are three variants of each CMSIS-NN function:
   *  - Functions without DSP Extension support (suitable for processors like Cortex-M0 and Cortex-M23, or a Cortex-M33 configured without the DSP Extension).
   *  - Functions with DSP Extension support (suitable for processors like Cortex-M4 and Cortex-M33 configured with the DSP Extension).
   *  - Functions that support both the DSP Extension and the MVE Extension (suitable for processors like Cortex-M55 configured with both the DSP Extension and the MVE Extension).
   * The right implementation is selected through compiler feature flags set by the user.

felix-johnny:
@EdmundPlayer Thanks for the feedback and apologies for the delay in addressing it. Please have  a look if it is better now.

Issue #620, opened 2019-07-02
Endless Loop in SCB_DisableDCache () with iMX-RT


Masmiseim36:
Hello,

with commit f46ec37262bc4fd82b6c91d24198cc3b4a499fe3 some ‘register’- qualifier are removed from the code. 

With this changes on NXP iMX-RT Controllers the Code execution enters an endless-loop when trying to disable the Data-Cache (--> calling the SCB_DisableDCache () function from the CMSIS Library) 
This problem is only valid when using memory, which is cached. When using the TCM only everything runs fine. Same with enabling the optimization – with optimization everything is fine. 
I crosschecked with STM32F7 and Kinetis KV58 to check if they have the same behavior. They work as intended. It looks like it is an iMX-RT exclusive problem.

The problem seems also to be valid only with GCC based Development Environment like McuXpresso, Rowley Crossworks and Segger EmbeddedStudio.

NXP has avoided the problem by using a quite old version of CMSIS in their SDK (commit 3960f196f93e695d6b8144328783e2db29d288fd).

It is curious that this problem seems to occur on iMXRT controller only, but not on others with the identical core. Could this be related to higher clock-speed – the iMXRT family has the highest M7-Clock I´m aware of – of cache-size – it is bigger than on the controllers I´ve compared with. Putting back the ‘register’- qualifier solves the problem for iMX-RT

Best Regards

Masmiseim36:
Little update:
The cache size shouldn´t be the reason as the iMXRT-1015 is also ending in an endless loop. This 1015 has also only 16 Kbyte Data and Instruction Cache like the STM32F7.

GuentherMartin:
Hello Masmiseim36,
can you please provide the Linker Description file or the the memory layout for your Sample-Code (CacheTest.zip).
I am trying to reproduce the problem using GCC.

Masmiseim36:
Hello GuentherMartin,

thanks for checking this issue.
Crossworks has an own XML-based Linker description which they translate automatically to a GCC-Linker File during building. I used the default one, which comes with the Development environment.
You can download an evaluation Version on their homepage: https://rowley.co.uk/
However, I added the generated Linker-Script to my post, which should work with the 1052-EVK-Board

In addition, I´ve added a sample project with MCUXpresso. Just Single-Step through the project to face the issue.

Thanks and best regards

[...]

PR #1426, opened 2022-02-26
Fix the endless loop issue with GCC O0.

Masmiseim36:
More details, see https://github.com/ARM-software/CMSIS_5/issues/620
The issue only happens when local variables are in stack (GCC O0). If local variables are saved
in general purpose register, then the function is OK.
When local variables are in stack, after disabling the cache, flush the local variables cache
line for data consistency.

### DIFF
--- a/CMSIS/Core/Include/cachel1_armv7.h
+++ b/CMSIS/Core/Include/cachel1_armv7.h
@@ -1,8 +1,8 @@
 /******************************************************************************
  * @file     cachel1_armv7.h
  * @brief    CMSIS Level 1 Cache API for Armv7-M and later
- * @version  V1.0.1
- * @date     19. April 2021
+ * @version  V1.0.2
+ * @date     21. June 2022
  ******************************************************************************/
 /*
  * Copyright (c) 2020-2021 Arm Limited. All rights reserved.
@@ -181,17 +181,21 @@ __STATIC_FORCEINLINE void SCB_EnableDCache (void)
 __STATIC_FORCEINLINE void SCB_DisableDCache (void)
 {
   #if defined (__DCACHE_PRESENT) && (__DCACHE_PRESENT == 1U)
-    uint32_t ccsidr;
-    uint32_t sets;
-    uint32_t ways;
+    struct {
+      uint32_t ccsidr;
+      uint32_t sets;
+      uint32_t ways;
+    } locals
+    #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
+       __attribute__((aligned(__SCB_DCACHE_LINE_SIZE)))
+    #endif
+    ;
 
     SCB->CSSELR = 0U;                       /* select Level 1 data cache */
     __DSB();
 
     SCB->CCR &= ~(uint32_t)SCB_CCR_DC_Msk;  /* disable D-Cache */
     __DSB();
-
-    ccsidr = SCB->CCSIDR;
 
     #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
       /*
@@ -206,24 +210,24 @@ __STATIC_FORCEINLINE void SCB_DisableDCache (void)
        * local variables cache line for data consistency.
        */
       /* Clean and invalidate the local variable cache. */
-      SCB->DCCIMVAC = (uint32_t)(&sets);
-      SCB->DCCIMVAC = (uint32_t)(&ways);
-      SCB->DCCIMVAC = (uint32_t)(&ccsidr);
+      SCB->DCCIMVAC = (uint32_t)&locals;
       __DSB();
       __ISB();
     #endif
+
+    locals.ccsidr = SCB->CCSIDR;
                                             /* clean & invalidate D-Cache */
-    sets = (uint32_t)(CCSIDR_SETS(ccsidr));
+    locals.sets = (uint32_t)(CCSIDR_SETS(locals.ccsidr));
     do {
-      ways = (uint32_t)(CCSIDR_WAYS(ccsidr));
+      locals.ways = (uint32_t)(CCSIDR_WAYS(locals.ccsidr));
       do {
-        SCB->DCCISW = (((sets << SCB_DCCISW_SET_Pos) & SCB_DCCISW_SET_Msk) |
-                       ((ways << SCB_DCCISW_WAY_Pos) & SCB_DCCISW_WAY_Msk)  );
+        SCB->DCCISW = (((locals.sets << SCB_DCCISW_SET_Pos) & SCB_DCCISW_SET_Msk) |
+                       ((locals.ways << SCB_DCCISW_WAY_Pos) & SCB_DCCISW_WAY_Msk)  );
         #if defined ( __CC_ARM )
           __schedule_barrier();
         #endif
-      } while (ways-- != 0U);
-    } while(sets-- != 0U);
+      } while (locals.ways-- != 0U);
+    } while(locals.sets-- != 0U);
 
     __DSB();
     __ISB();
--- a/CMSIS/Core/Include/cachel1_armv7.h
+++ b/CMSIS/Core/Include/cachel1_armv7.h
@@ -2,7 +2,7 @@
  * @file     cachel1_armv7.h
  * @brief    CMSIS Level 1 Cache API for Armv7-M and later
  * @version  V1.0.2
- * @date     21. June 2022
+ * @date     22. June 2022
  ******************************************************************************/
 /*
  * Copyright (c) 2020-2021 Arm Limited. All rights reserved.
@@ -186,8 +186,8 @@ __STATIC_FORCEINLINE void SCB_DisableDCache (void)
       uint32_t sets;
       uint32_t ways;
     } locals
-    #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
-       __attribute__((aligned(__SCB_DCACHE_LINE_SIZE)))
+    #if ((defined(__GNUC__) || defined(__clang__)) && !defined(__OPTIMIZE__))
+       __ALIGNED(__SCB_DCACHE_LINE_SIZE)
     #endif
     ;
 
@@ -197,12 +197,12 @@ __STATIC_FORCEINLINE void SCB_DisableDCache (void)
     SCB->CCR &= ~(uint32_t)SCB_CCR_DC_Msk;  /* disable D-Cache */
     __DSB();
 
-    #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
+    #if ((defined(__GNUC__) || defined(__clang__)) && !defined(__OPTIMIZE__))
       /*
-       * For the endless loop issue with GCC O0.
+       * For the endless loop issue with GCC and clang with O0.
        * More details, see https://github.com/ARM-software/CMSIS_5/issues/620
        *
-       * The issue only happens when local variables are in stack (GCC O0). If
+       * The issue only happens when local variables are in stack (GCC/clang O0). If
        * local variables are saved in general purpose register, then the function
        * is OK.
        *
--- a/CMSIS/Core/Include/cachel1_armv7.h
+++ b/CMSIS/Core/Include/cachel1_armv7.h
@@ -193,6 +193,25 @@ __STATIC_FORCEINLINE void SCB_DisableDCache (void)
 
     ccsidr = SCB->CCSIDR;
 
+    #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
+      /*
+       * For the endless loop issue with GCC O0.
+       * More details, see https://github.com/ARM-software/CMSIS_5/issues/620
+       *
+       * The issue only happens when local variables are in stack (GCC O0). If
+       * local variables are saved in general purpose register, then the function
+       * is OK.
+       *
+       * When local variables are in stack, after disabling the cache, flush the
+       * local variables cache line for data consistency.
+       */
+      /* Clean and invalidate the local variable cache. */
+      SCB->DCCIMVAC = (uint32_t)(&sets);
+      SCB->DCCIMVAC = (uint32_t)(&ways);
+      SCB->DCCIMVAC = (uint32_t)(&ccsidr);
+      __DSB();
+      __ISB();
+    #endif
                                             /* clean & invalidate D-Cache */
     sets = (uint32_t)(CCSIDR_SETS(ccsidr));
     do {


CMSIS/Core/Include/cachel1_armv7.h
```diff
@@ -193,6 +193,25 @@ __STATIC_FORCEINLINE void SCB_DisableDCache (void)
...
+       * local variables are saved in general purpose register, then the function
+       * is OK.
+       *
+       * When local variables are in stack, after disabling the cache, flush the
+       * local variables cache line for data consistency.
+       */
+      /* Clean and invalidate the local variable cache. */
+      SCB->DCCIMVAC = (uint32_t)(&sets);
+      SCB->DCCIMVAC = (uint32_t)(&ways);
+      SCB->DCCIMVAC = (uint32_t)(&ccsidr);
```
flit: Performing three cache line flushes is a bit overkill. It would be better to place the local variables in a struct to ensure co-location, then only perform a cache line flush on that struct's address. Note the necessity to align the struct to the D$ line size.

```c
struct {
    uint32_t ccsidr;
    uint32_t sets;
    uint32_t ways;
} locals __attribute__((aligned(__SCB_DCACHE_LINE_SIZE));

SCB->DCCIMVAC = (uint32_t)&locals;
__DSB();
__ISB();
```

flit (changes requested):

CMSIS/Core/Include/cachel1_armv7.h
Masmiseim36: Agree, three cache line flushes in a row are definitely an overkill, but it simplifies the rest of the code.

flit: True, it does make it more complex for sure. But it's architecturally cleaner. So thanks for the update!

Btw, you will need to bump the version (to 1.0.2) and date at the top of the file to get the file header check to pass.

CMSIS/Core/Include/cachel1_armv7.h
```diff
@@ -181,30 +181,53 @@ __STATIC_FORCEINLINE void SCB_EnableDCache (void)
...
-    uint32_t ccsidr;
-    uint32_t sets;
-    uint32_t ways;
+    struct {
+      uint32_t ccsidr;
+      uint32_t sets;
+      uint32_t ways;
+    } locals
+    #if (defined(__GNUC__) && !defined(__OPTIMIZE__))
+       __attribute__((aligned(__SCB_DCACHE_LINE_SIZE)))
```
flit: This attribute will have to be replaced with `__ALIGNED(__SCB_DCACHE_LINE_SIZE)` from the CMSIS compiler polyfill definitions (`cmsis_compiler.h` and its includes). Even though it's currently only for gcc, it should match the rest of CMSIS. Sorry, I should have used it in my example, but was just writing it quickly in the browser…

Speaking of, does this issue really only affect gcc? What about clang, armclang, IAR, etc?

Masmiseim36: IAR and Keil are not affected. When disabling optimization on those compilers they still do some minor optimization which leads to placing Variables in registers only instead of using the stack.
Disabling optimization on the GCC means that there is really no optimization, and the variables are placed on the stack. I’ve done some tests with clang and it behaves like GCC.
I’ve updated the PR to fix also the clang issue. The __ALIGNED macro is also used.

JonatanAntoni:
@grasci-arm test this please

CMSIS/Core/Include/cachel1_armv7.h
flit: Thanks for the updates!

flit (approved):

TTornblom:
Is the stack alignment to the cache line size really necessary in commit https://github.com/ARM-software/CMSIS_5/pull/1426/commits/30708866745f5430917b1706181cd5d9a1eb429f ?

Non-optimized builds with the IAR compiler also suffers from the endless loop issue, but the stack can not be aligned to match the cache line size due to the way the stack works in IAR builds.
Looking at the code I don't see how this would be necessary either. None of the variables are accessed before the cache has been disabled, so if I understand it correctly the variables in the struct will never enter the cache and thus there is no need to align them on a cache line.

Is my understanding correct or not?

Wastus:
Is there a reason why these fixes are only applied to one location `SCB_DisableDCache` and not the other functions which basically use a clone of the code? I suffered from this using `SCB_CleanDCache`.

JonatanAntoni:
Hi @Wastus,

This was a community contribution from @Masmiseim36.

Can you check if the same change would solve your issue in `SCB_CleanDCache`?
Cache handling is a complex procedure and unfortunately it seems to be very hard to have this implemented in compiler agnostic C code.

As a last resort, I could only propose to implement cache maintenances routines in assembly specific to your requirements. You could use disassembled code using high optimization as a starting point.

Masmiseim36:
Hallo @Wastus,
Hello @JonatanAntoni,

Cache stuff is a quite complicated thing. 
It is not always easy to create a reproducible scenario that works for all supported compilers. Unfortunately, these often behave somewhat differently. In my tests it was not necessary to introduce this special treatment for IAR and Keil, but it was for GCC and CLANG.

In PR #1611, the flush for the local variables was also extended for IAR and Keil. Probably the behavior of both compilers has changed in the meantime, maybe my test was not correct.

Because of the complexity and because my first request in this direction (#620) was rejected with reference to a compiler bug, I had concentrated in this PR on the functionality that I could test well and where I had problems in my application.

njankowski-renesas:
Hi All,

Expanding on this topic with some exposition.

### Problem One
The problem with `SCB_DisableDCache` that this PR and the companion IAR/Keil PR mitigates is an edge case with Cortex-M7 cache behavior.

**For the Cortex-M7, if `CCR.DC` is `0`, reading or writing locations that are still dirty in the cache must be prohibited until the location has been cleaned from the cache.**

Reading locations that are still dirty in the cache while `CCR.DC` is `0`, will result in stale data being read from external memory.

Writing locations that are still dirty in the cache while `CCR.DC` is `0`, and then cleaning or cleaning and invalidating those same modified locations will overwrite the locations with stale data from the cache.

#### Further Explanation
For this issue with unoptimized builds using stack variables, a window of opportunity exists in `SCB_DisableDCache` where `CCR.DC` is `0` but dirty cache lines are still present which includes the stack variables, and the stack variable locations are written before being cleaned from the data cache.

It is circumstantial that the stack is affected, as any dirty cached location could be affected even in optimized builds.

This means that even if registers were used for the variables, or even with this mitigation for unoptimized builds, there are still issues that may occur.

For example, if an exception is taken before the clean and invalidate operation in `SCB_DisableDCache` completes, if the exception handler accesses dirty locations while the clean and invalidate sequence has not been completed, that is a problem. Also consider a context switch in an RTOS environment.

The stack variables could also feasibly be in a non-cacheable location, and this specific issue would not be seen.

The problem is situational.

### Problem Two
The problem of stack variable usage with `SCB_InvalidateDCache` and `SCB_InvalidateDCache_by_Addr` remains unmitigated in CMSIS 5 or 6 with unoptimized builds using stack variables.

**For Cortex-M7, Cortex-M55, and Cortex-M85, the `SCB_InvalidateDCache` and `SCB_InvalidateDCache_by_Addr` functions may invalidate their own stack variables if stack variables are used.**

Invalidation destroys cache lines whether they are dirty or clean.

If the stack variables are cached, they will be destroyed by full invalidation or may be partially or fully destroyed by address-based invalidation.

This has nothing to do with the `CCR.DC` setting and does not represent a cache behavior edge case in any of the listed Cortex-M.

This PR and the companion IAR/Keil PR do not mitigate the issue with these functions.

#### Further Explanation

The nature of invalidation creates this problem.

Also, consider that if an exception is taken before the invalidate operation in `SCB_InvalidateDCache` or `SCB_InvalidateDCache_by_Addr` completes, if the exception handler depends on the invalidation being completed, that is a problem. Also consider a context switch in an RTOS environment.

The stack variables could feasibly be in a non-cacheable location, may be write-through, or may be spared destruction by an address-based invalidation, and this specific issue would not be seen.

The problem is situational.

---

### Why Problem One is CM7 Only
The newer Cortex-M55 and Cortex-M85 should not have this issue because of new cache behaviors provided by the `MSCR.xCACTIVE` bits, which do not exist for Cortex-M7.

For the specific example of `SCB_DisableDCache` in an unoptimized build using stack variables:

For CM7, because both cache allocations and cache lookups stop when `CCR.DC=0`, the dirty cached data for the stack variables has no way to become coherent when writes to the stack variable locations occur after `CCR.DC` has been set to `0`.

A clean and invalidate will thus write the stack variable locations with the dirty cached data, now outdated, which they had before `CCR.DC` was set to `0`, when they are finally cleaned from the data cache.

For CM55 and CM85, because only cache allocations stop when `CCR.DC=0` but `MSCR.DCACTIVE=1` allows lookups to continue, the existing cached data for the stack variables will be updated by writes to the stack variable locations even after `CCR.DC` has been set to `0`.

A clean and invalidate will thus write the stack variable locations with the correct dirty cached data, when they are finally cleaned from the data cache.

This no-allocation, but continued lookup cache behavior solves the edge case for read and write access to dirty locations when `CCR.DC=0` and the cache is being cleaned and invalidated.

---

### Recent No-Optimization Compilation Results
I checked the disassembly on CMSIS 5.9.0 (prior to this patch) on the following latest compilers and built for a Cortex-M85 target, and all generated similar stack access code in the affected functions that would cause an issue when their optimization was turned off.
- Arm GNU Toolchain Version 13.2.Rel1 (GCC)
- LLVM Embedded Toolchain for Arm 17.0.1 (Clang)
- Arm Compiler for Embedded 6.21 (AC6, LLVM based)

The latest IAR did not seem to generate stack access code and was OK.
- IAR Build Tools for Arm 9.50.1

I assume similar code would be generated for a Cortex-M7 target.

Snippet of `SCB_DisableDCache` from LLVM with no optimization `-O0` that shows stack access:
```
SCB->CCR &= ~(uint32_t)SCB_CCR_DC_Msk;
    2001552     6808       ldr r0, [r1, #0]
    2001554     f420 3080  bic.w r0, r0, #65536 @ 0x10000
    2001558     6008       str r0, [r1, #0]
    200155a     f3bf 8f4f  dsb sy
    200155e     f64e 5080  movw r0, #60800 @ 0xed80
    2001562     f2ce 0000  movt r0, #57344 @ 0xe000
ccsidr = SCB->CCSIDR;
    2001566     6800       ldr r0, [r0, #0]
    2001568     900d       str r0, [sp, #52] @ 0x34
sets = (uint32_t)(CCSIDR_SETS(ccsidr));
    200156a     980d       ldr r0, [sp, #52] @ 0x34
    200156c     f3c0 304e  ubfx r0, r0, #13, #15
    2001570     900c       str r0, [sp, #48] @ 0x30
...
These load-store sequences that access the stack are the problem.
```

---

### Questions
@Masmiseim36

Because I have not tested the Cortex-M7 issue myself, was this truly an "infinite loop" or did the stack variables simply become really large from stale data by coincidence?

In theory the stack variables could be overwritten to any value, meaning this issue could have gone unnoticed as an "infinite loop" if the values were overwritten to small values and caused the loop to terminate early, leaving dirty lines in the data cache.

@Wastus

I do not see how this could occur for `SCB_CleanDCache`, since the stack variables should remain coherent with `CCR.DC` never being set to `0` in the function.

The problem with `SCB_DisableDCache` is that it breaks the first above bolded rule when a compiler generates writes to locations that are still dirty in the cache after cache lookups have been disabled, but before the dirty locations have been cleaned from the cache.

---

### Data Cache Functions
| Function | Issue | Affects |
| - | - | - |
| SCB_EnableDCache | None | None |
| SCB_DisableDCache | Problem one. `CCR.DC=0` with dirty cache lines precludes read or write access to the dirty locations until the locations are cleaned, which may include the stack variables if they are present, cached, and dirty. | CM7 |
| SCB_InvalidateDCache | Problem two. Invalidation destroys all cache lines dirty or clean, which may include the stack variables if they are present, cached, and not write-through. | CM7, CM55, CM85 |
| SCB_CleanDCache | None | None |
| SCB_CleanInvalidateDCache | None | None |
| SCB_InvalidateDCache_by_Addr | Problem two. Invalidation destroys all cache lines dirty or clean, which may include the stack variables if they are present, cached, not write-through, and if the target address and length invalidates the containing cache line(s). | CM7, CM55, CM85 |
| SCB_CleanDCache_by_Addr | None | None |
| SCB_CleanInvalidateDCache_by_Addr | None | None |

I assume the hardware will properly hazard read and write access to locations where an invalidate, clean, or clean and invalidate cache maintenance operation is currently taking place, so long as cache lookup is enabled for the respective Cortex-M type.

An application still has to be responsible for knowing if interrupting cache maintenance in any of these functions will be a problem for the application logic.

Generally, the data cache invalidate-only functions should not be used and should be replaced with clean-and-invalidate functions, since they eliminate the data loss risk with cache lines that accidentally mix dirty locations and locations to be invalidated on the same cache line(s).

### Cache Behavior
| Core | CCR.xC | MSCR.xCACTIVE | Behavior |
| - | - | - | - |
| CM7 | 1 | Not Available | Allocate, Lookup |
| CM7 | 0 | Not Available | No Allocate, No Lookup |
| CM55, CM85 | 1 | 1 | Allocate, Lookup |
| CM55, CM85 | 0 | 1 | No Allocate, Lookup |
| CM55, CM85 | X | 0 | No Allocate, No Lookup |

### References
For Cortex-M7, reference `5.9 L1 caches` in the latest Technical Reference Manual (TRM).

```
If the access is to Non-shared cacheable memory, and the cache is enabled, a lookup is performed in the cache and, if found in the cache, that is, a cache hit, the data is fetched from or written into the cache.
When the cache is not enabled and for Non-cacheable or Shared memory the accesses are performed using the AXIM interface.
```

For Cortex-M55 and Cortex-M85, reference `10.9.6 Accessing the caches` and `Table 10-13: Cache access scenarios` in their latest TRMs.
| CCR | MSCR | Cache access behavior |
| - | - | - |
| CCR.DC and CCR.IC are set to 1 | MSCR.DCACTIVE and MSCR.ICACTIVE are set to 1 | Normal operating mode. Unless PDCORE goes OFF resulting in PDRAMS going to RET, the caches are powered up and cache accesses can perform allocation and lookup. |
| CCR.DC and CCR.IC are set to 0 | MSCR.DCACTIVE and MSCR.ICACTIVE are set to 1 | Cache lookups are allowed, but cache allocation is not permitted. This behavior is used to clean the cache before powering down. |
| CCR.DC and CCR.IC are set to 0 or 1 | MSCR.DCACTIVE and MSCR.ICACTIVE are set to 0 | The caches are not being used, and they can be powered down. The CCR.DC and CCR.IC bits are ignored. |

Masmiseim36:
Hello @njankowski-renesas 

Thank you for your detailed explanation. The whole cache-topic is really complicated. 

According to my observation, the following happens without an invalidate cache of the local variables:

Precondition:
- There is no optimization which means that the local variables are on the stack
- The currently used area of the stack is in the cache (which is quite likely)

Process:
The first loop passes are executed as expected. The cache is cleand and invalidated based on its ways. Until the way containing the stack is to be invalidated. The clean/invalidate of this way sets the value of the local variable to an invalid (too large) value.

Incidentally, my original statement that this is an infinite loop was incorrect. The loop is finite, but this can take up to ten minutes on a fast microcontroller.


If you clear/invalidate the stack before the clean/invalidate loop, you can avoid this error.  In my opinion, this is a valid solution.

Regards

njankowski-renesas:
Hi @Masmiseim36,

Thanks for your reply!

> without an invalidate cache of the local variables

Clean and invalidate, that is :)

> Incidentally, my original statement that this is an infinite loop was incorrect. The loop is finite, but this can take up to ten minutes on a fast microcontroller.

This is great to hear, because it helps validate my thought experiment on the issue.

> If you clear/invalidate the stack before the clean/invalidate loop, you can avoid this error. In my opinion, this is a valid solution.

Agreed.
This is a valid mitigation for the first problem.

Unfortunately the invalidate-only functions for data cache are also affected as I described for the second problem, but there is no mitigation that can be done without breaking the expectations of the functions.
If this clean and invalidate mitigation were to be used for the invalidate-only functions, the stack variables would have to be the only data that is cleaned, meaning they would need to occupy a cache line exclusively.
Really, using a clean cache maintenance operation anywhere inside of the invalidate-only functions is antithetical to their expected behavior, so I would say such workaround is not acceptable for problem two.

The fortunate aspect is I would assume running into this scenario with the invalidate-only functions would be unlikely, since they are probably seldom used by most applications, or are used in ways where this problem would not occur.

Regardless, problem two persists as an issue to be aware of.

PR #1434, opened 2022-03-11
Corrects spelling

Alex-EEE:


grasci-arm:

jkrech:
@Alex-EEE thanks for your proposal, however if we would accept your change this would mean that we have the same symbol defined twice.
The background why we have duplicated the Symbol for value 11 is the fact that we had this type LENGHT vs. LENGTH in our initial version and the misspelled symbol got used in Driver Implementations. In order to correct the symbol without breaking the build of existing drivers, we left the symbol with the wrong spelling and marked it as deprecated in the comment.

Therefore I am asking you to close your PR without your code being merged.

Thanks a lot.

~ Joachim

Alex-EEE:
Oh great, I see, you already fixed it, great!

PR #1435, opened 2022-03-12
RTX5: Remove unused member of osRtxThread_t

shosakam:
tz_memory has never been used when DOMAIN_NS is enabled.


grasci-arm:

RobertRostohar:
It is correct that 'tz_memory' member in the Thread Control Block structure is only used when DOMAIN_NS is enabled.

However it is not desired to change the size of the Thread Control Block. That would break debuggers which show thread information and expect a fixed Thread Control Block.

Therefore the compromise is to leave that member in the structure.

Note that the other define RTX_TF_M_EXTENSION is more experimental and is treated different.

shosakam:
Thank you for your explanation. I understand.

PR #1443, opened 2022-03-17
DSP: Remove undefined behavior in post-increment/decrement read

mortior:
According to the C standard q15_t * and const q15_t * are not compatible
types which, among other things, imply that an object of type const
q15_t * can't be modified by writing to a q15_t ** or vice versa.
Programs doing this are undefined.

Because of this rule all programs using the functions read_q15x2_ia,
read_q15x2_da, read_q7x4_ia, or read_q7x4_da for reading data from an
array of constant elements will be undefined. To solve this it is not
enough to change the type of the function since this will give problems
when reading data from an array of non-const elements. To get a defined
solution I needed to switch from functions to macros to allow the
increment to be done in the original type of the pointer.

### DIFF
--- a/CMSIS/DSP/Include/arm_math_memory.h
+++ b/CMSIS/DSP/Include/arm_math_memory.h
@@ -74,7 +74,7 @@ extern "C"
   @return        Q31 value
  */
 __STATIC_FORCEINLINE q31_t read_q15x2 (
-  q15_t * pQ15)
+  q15_t const * pQ15)
 {
   q31_t val;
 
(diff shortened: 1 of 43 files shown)


christophe0606:
@mortior Hi.

I am, not sure I understand what's your issue here.
Of course, modifying a const array through a normal pointer is undefined and a bug. But (I hope) we don't do this in the library.

What we are doing is to remove the const qualifier but we do not change the content of the array.
And this is totally acceptable thing in C.

It is even done in the C standard library) : for instance https://github.com/eblot/newlib/blob/master/newlib/libc/string/memchr.c

So I don't see any problem with functions like read_q15x2_ia.
We are not changing the content of the pointed const array. The pointer is just incremented.

mortior:
@christophe0606  Hi,
The problem is not with the array itself is with the pointer (or cursor as I call it). According to the aliasing rules in the C standard writing through a dereferenced pointer of type `q15_t **` can't change an object of type `const q15_t *` and trying to do this invokes undefined behavior. This implies that all functions that takes an object of type `const q15_t *` (e.g `const q15_t * src`) takes its address and passes that address to `read_q15x2_ia` (e.g. `read_q15x2_ia((q15_t**)&src)`) will invoke undefined behavior. 

Note that `const int **` is not a const-qualified `int **` with respect to the C-standard, the const-qualified version of `int **` is `int * const *`.

Changing the declaration to `read_q15x2_ia(const int **)` will not help since sometimes the cursor is of type `q15_t *` and casting the address of such a cursor to `const q15_t **` will suffer from the same undefined behavior. This is the reason for the somewhat complicated macros used for implementing `read_q15x2_ia` and friends.

christophe0606:
@mortior Thank you for this PR. It forced me to have a very interesting discussion with our compiler developers and now I understand what the problem is.
This PR is impacting lot of files so from a validation point of view it will require more effort from me but we have a release coming soon.

So, I'll merge this PR only after the release. 

Also I still need to think about the macros and if I can do differently (avoiding macros ... but I don't think it is possible).

mortior:
@christophe0606 
As input on when you want merge it I can add that the old code breaks on the IAR Arm compiler (which is why it ended up on my table in the first place).

I have tried writing it without macros but didn't succeed, all versions I came up with contained the same UB as the original code.

christophe0606:
@mortior I am generally quick on CMSIS-DSP support but unfortunately I have been asked to work on a priority issue and I can't look at it before end of April.

Of course, if the situation is changing in the meantime I'll come back to this as soon as I can.

Sorry for the delay.

christophe0606:
@mortior I have tested and merged it.

PR #1448, opened 2022-03-29
Update core_starmc1.h

AlbertHuang-CPU:
Add macro define of __STAR_MC


grasci-arm:

JonatanAntoni:
@AlbertHuang-CPU, please fix the warnings. CMSIS-Headers are often used by copy from the development branch. In order to keep track we have version and date information in each file which need to be updated on every change.

PR #1450, opened 2022-03-29
Fixing compilation issue with Visual Studio 2022 in Arm64EC

jbr-smtg:
Dear @christophe0606, 

The goal of this pull request is to fix a compilation issue encountered with Visual Studio 2022 if the Arm64EC mode is used.
For some reasons, including the <arm_neon.h> directly in Arm/Arm64 works, but not in Arm64EC. As a workaround it is possible to use <arm64_neon.h> instead in Arm64EC.

I don't know if there is a particular reason for Microsoft to do that (I have asked here: https://docs.microsoft.com/en-us/answers/questions/791892/why-is-it-not-working-to-include-arm-neonh-in-a-ar.html). This might also just be a bug on Microsoft side?

Meanwhile for CMSIS to support Visual Studio 2022 with Arm64EC, the following change is required.

Thanks in advance!

Best Regards,
JbR

### DIFF
--- a/CMSIS/DSP/Include/arm_math_types.h
+++ b/CMSIS/DSP/Include/arm_math_types.h
@@ -95,12 +95,16 @@ extern "C"
 #endif
 
 #if defined(ARM_MATH_NEON)
-#include <arm_neon.h>
-#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
-  #if !defined(ARM_MATH_NEON_FLOAT16)
-  #define ARM_MATH_NEON_FLOAT16
+  #if defined(_MSC_VER) && defined(_M_ARM64EC)
+    #include <arm64_neon.h>
+  #else
+    #include <arm_neon.h>
   #endif
-#endif
+  #if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC) && __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
+    #if !defined(ARM_MATH_NEON_FLOAT16)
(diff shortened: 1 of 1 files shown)


christophe0606:
@jbr-smtg Sorry for the late answer. I was out of office.
I'll ask our compiler people before I merge this PR.

sjoerdmeijer:
Hello guys. I don't know much about Arm64EC, but what is _arm64_neon.h_? I don't see this being defined in the ACLE.

sjoerdmeijer:
A quick google shows:

> NEON intrinsics are supported, as provided in the header file arm64_neon.h. The MSVC support for NEON intrinsics resembles that of the ARM64 compiler, which is documented in the [ARM NEON Intrinsic Reference](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics) on the ARM Infocenter website.

Copied from: https://docs.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170

So if it can be confirmed that arm64_neon.h is an officially supported SDK header, then this looks fine I guess.

jbr-smtg:
@sjoerdmeijer From my understanding <arm64_neon.h> seems to be Microsoft specific. If Microsoft should not do it this way, it would be great to have an official statement that we could communicate to them?
Otherwise I don't see any other solution than adding a Visual Studio specific workaround... Including this file or defining the _M_HYBRID_X86_ARM64 macro before including <arm_neon.h>.

christophe0606:
@jbr-smtg The current fix looks ok to me. I'll merge it as it is.

sjoerdmeijer:
Agreed, looks fine since this is specifically checking for _MSC_VER _M_ARM64EC.

jbr-smtg:
Many thanks :)

Issue #1472, opened 2022-04-28
CMSIS-NN: Bug in SVDF when ARM_MATH_DSP is defined


hmogensen-arm:
Blockwise matmul yields incorrect output

PR #1473, opened 2022-04-28
CMSIS-NN: Correct SVDF bug for ARM_MATH_DSP #1472

hmogensen-arm:
Fix bug in SVDF function with ARM_MATH_DSP define
Fix for: #1472


grasci-arm:

PR #1477, opened 2022-05-05
DSP: Fix uninitialised variable references

stephanosio:
This commit adds initialisation for the variables that can be used
potentially uninitialised so that GCC does not generate warnings while
building the MVE function implementations.

Signed-off-by: Stephanos Ioannidis <root@stephanos.io>


stephanosio:
cc @christophe0606

grasci-arm:

PR #1479, opened 2022-05-06
CMSIS-NN: Initialize unused bias dims in unit tests to avoid warnings

mansnils:
Avoid warnings that may turn into errors.


grasci-arm:

felix-johnny (approved):

PR #1481, opened 2022-05-10
CMSIS-NN: Disable loop unroll for depthwise_conv_s8_mult_4 for GCC

mansnils:


grasci-arm:

PR #1494, opened 2022-05-25
Suggestion to add exact sum functions in BasicMathFunctions folder 

Silfurion:
Hello,
In some of our projects, we would like to accelerate the sum of vector elements. At the moment CMSIS doesn't offer a simple sum function. The closest method was the mean method that unfortunately doesn't offer an exact result if we want to multiply its result by the size of the array to get back the sum of the vector.

With this pull request we would like to suggest the addition of a new sum function that would allow to benefit from the already implemented and accelerated mean function, but without the final division to obtain a more accurate sum result.

We added these function in the BasicMathFunctions folder available for q7 , q15 , q31 , float16 , float32 and float64.

Thanks again


grasci-arm:

christophe0606:
@Silfurion I'll need to write and add some tests before I can merge this PR.
Also, I'll rename the function to accumulate if you don't mind.

grasci-arm:

christophe0606:
There were some pending questions in this PR.

Accumulation functions are now tracked in this new github issues on the new CMSIS-DSP repository : https://github.com/ARM-software/CMSIS-DSP/issues/12

PR #1497, opened 2022-05-27
Fix build errors in arm_bitreversal2.S on CM23

SeanMeng0509:
'IT' instruction is not supported by CM0 family and CM23.


grasci-arm:

SeanMeng0509:
Wrong target branch

PR #1498, opened 2022-05-27
Fix build errors in arm_bitreversal2.S on CM23

SeanMeng0509:
'IT' instruction is not supported by CM0 family and CM23.


grasci-arm:

christophe0606:
@SeanMeng0509 Do you have any use of "arm_bitreversal2.S" ? Because on recent versions of CMSIS-DSP it is no more used and instead there is a C file implementation.
This .S file should have been removed. It is no more included by our build (CMSIS Pack or cmake).

grasci-arm:

christophe0606:
File removed from the repository since it was no more used.

PR #1502, opened 2022-06-13
SVD: fix regex of bitRangeType

duskmoon314:
Fix conflict between the comment and the regex

### DIFF
--- a/CMSIS/Utilities/CMSIS-SVD.xsd
+++ b/CMSIS/Utilities/CMSIS-SVD.xsd
@@ -253,7 +253,7 @@
   <!-- bitRangeType specifies the bit numbers to be restricted values from 0 - 69 -->
   <xs:simpleType name="bitRangeType">
     <xs:restriction base="xs:token">
-      <xs:pattern value="\[([0-4])?[0-9]:([0-4])?[0-9]\]"/>
+      <xs:pattern value="\[([0-6])?[0-9]:([0-6])?[0-9]\]"/>
     </xs:restriction>
   </xs:simpleType>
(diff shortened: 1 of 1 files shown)


jkrech:
Given the fact that today SVDConv is limited to a maximum size of a register to 32-bit, the range value should be 0 to 31, despite the comment reading 69 which I cannot make sense of right now. Am I missing something?

duskmoon314:
> despite the comment reading 69 which I cannot make sense of right now

I think this might be 0-63 which means a 64-bit register. And maybe the first writer thinks it is easy to build up a regex for 69. (I cannot find previous commits on this line in the repo)

I am trying to build up an SVD file of a RISC-V MCU and it has a peripheral with three 64-bit registers. So I see the error of `bitRange`, finding this line and the comment, then open this PR.

grasci-arm:

JonatanAntoni:
Please, could you raise this PR to the new location of SVD spec?
Find the discussed spec at https://github.com/Open-CMSIS-Pack/svd-spec/blob/12dc76458ea68e9c1a7b81689207216c1fefa053/schema/CMSIS-SVD.xsd#L256

duskmoon314:
> Please, could you raise this PR to the new location of SVD spec? Find the discussed spec at [Open-CMSIS-Pack/svd-spec@`12dc764`/schema/CMSIS-SVD.xsd#L256](https://github.com/Open-CMSIS-Pack/svd-spec/blob/12dc76458ea68e9c1a7b81689207216c1fefa053/schema/CMSIS-SVD.xsd#L256)

Sure!

PR #1506, opened 2022-06-17
CMSIS_DSP: arm_fir_example accessing past the end of firStateF32

jritchey-renesas:
When built with `ARM_MATH_MVEF`, `arm_fir_f32()` is accessing 3 elements past the end of `firStateF32` in the arm_fir_example.

### DIFF
--- a/CMSIS/DSP/Examples/ARM/arm_fir_example/arm_fir_example_f32.c
+++ b/CMSIS/DSP/Examples/ARM/arm_fir_example/arm_fir_example_f32.c
@@ -171,9 +171,9 @@ static float32_t testOutput[TEST_LENGTH_SAMPLES];
  * Declare State buffer of size (numTaps + blockSize - 1)
  * ------------------------------------------------------------------- */
 #if defined(ARM_MATH_MVEF) && !defined(ARM_MATH_AUTOVECTORIZE)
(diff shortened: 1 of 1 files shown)


grasci-arm:

christophe0606:
@jritchey-renesas

You're right. The example has not been updated. For performance reason, the state with MVEF is bigger and saving more data.

When I am looking at the `arm_fir_init_f32`, I can see

```c
#if defined(ARM_MATH_MVEF) && !defined(ARM_MATH_AUTOVECTORIZE)
  memset(pState, 0, (numTaps + (blockSize - 1U) + blockSize) * sizeof(float32_t));
#else
  memset(pState, 0, (numTaps + (blockSize - 1U)) * sizeof(float32_t));
#endif
```

Note that for `MVEF` there is another constraint documented in the release notes:

`For performance reasons, some vectorized algorithms may read one vector after the end of buffer (but the data read is not used). If the buffer is close to the end of a memory zone, it is advised to leave some free memory area between the end of the buffer and the end of the memory zone.`

But this is a constraint on the input buffer and not the state.

We are in the process of moving the CMSIS-DSP to a new github repository so I'll open a github issue in the new repository to remember to correct this example.

christophe0606:
Now tracked in another github issue. This will require some investigation to correct the problem.

PR #1512, opened 2022-06-22
CMSIS-NN: Add batch support for 1xN conv

felix-johnny:
Unit tests are updated as well.

Change-Id: If0241f3311756cd4548c4ae3c43573f8f43dfa3d


grasci-arm:

CMSIS/NN/Tests/UnitTest/generate_test_data.py
```diff
@@ -38,6 +38,7 @@
 
 REQUIRED_MINIMUM_TENSORFLOW_VERSION = version.parse("2.5")
 TESTDATA_SETS = {}
+ALL_TESTDATA_SETS = {}
```
mansnils: I think this should be removed

felix-johnny: Without that I get a 'NameError: name 'ALL_TESTDATA_SETS' is not defined' error when generating data..For e.g './generate_test_data.py --dataset kernel1x1 -a'

mansnils: Ah right, it seems to be the result of a merge issue. So the current ALL_TESTDATA_SETS should be replaced by just TESTDATA_SETS.

felix-johnny: ok..i'll fix that..thanks..!

CMSIS/NN/README.md
```diff
@@ -28,7 +28,7 @@ Group | API | Base Operator | Input Constraints | Additional memory required for
 ||arm_convolve_wrapper_s8()|CONV| None |n.a.| Yes | Yes |The additional memory required depends on the optimal convolution function called.|
 ||arm_convolve_s8()|CONV| None |4 * (ker_x * ker_y * input_ch + delta)| Yes | Yes |delta - MVE only|
```
mansnils, line 29: Looks good..

mansnils (approved):

PR #1521, opened 2022-07-08
CMSIS-NN: Add MVEI support for int16 depth-wise convolution

hmogensen-arm:


grasci-arm:

PR #1523, opened 2022-07-12
Fix undefined identifier in system_x headers

fkjagodzinski:
All header files present in `Device/ARM/x/Include/system_x.h` path declare `SystemCoreClock` as `uint32_t` but some of them fail to provide a typedef for `uint32_t`. As a result, one can only use these headers after providing a definition of the type.

Add a missing include directive for `stdint.h` and fix the following error:

> Device/ARM/ARMCM85/Include/system_ARMCM85.h:40:8: error:unknown type name 'uint32_t'
>    40 | extern uint32_t SystemCoreClock;


grasci-arm:

PR #1526, opened 2022-07-18
CMSIS-NN: Improve MVE implementation of average_pool

felix-johnny:
The s8 implementation is improved to use more vector instructions
and the variable naming is changed to be consistent with other
parts of CMSIS-NN.
Tail handling is folded into the main loop by using predicated instructions.
Test case is updated to go inside the main loop..

Change-Id: I7e2ccba9f4081e56597e8bc9311037e170f91baa


grasci-arm:

PR #1541, opened 2022-08-05
CMSIS-NN Cleanup Doxygen comments

felix-johnny:
Duplicate function descriptions  are removed. The header
is retained to be the main reference.


grasci-arm:

mansnils (approved):

PR #1545, opened 2022-08-09
CMSIS-NN: Block process for DW conv

felix-johnny:
DW conv for MVE is changed to process in smaller
blocks. For most cases this results in lower scratch
buffer requirement.


grasci-arm:

mansnils (approved):

PR #1548, opened 2022-08-11
CMSIS-NN : Clear output buffer in unit tests if reused

### LINKED ISSUE #1532: CMSIS-NN - Overwrite `output` Between Consecutive Unit Tests
The CMSIS-NN unit tests often consecutively execute functions in the same unit test case. They make use of the same input values and parameters and are expected to produce the same `output`. Taking the [`test_arm_convolve_s8.c`](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_s8/test_arm_convolve_s8.c) as an example. Here, we first test [`arm_convolve_s8()`](https://github.com/ARM-software/CMSIS_5/blob/0a42364521c589690ffc5c63bdf662c05bd0f27a/CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_s8/test_arm_convolve_s8.c#L88) and then [`arm_convolve_wrapper_s8`](https://github.com/ARM-software/CMSIS_5/blob/0a42364521c589690ffc5c63bdf662c05bd0f27a/CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_s8/test_arm_convolve_s8.c#L88) in every test case; using the same input and `output` variables. 

While tinkering around, I came across the (admittedly) rare case that a modified (broken) version of `arm_convolve_wrapper_s8()` would not modify the `output` array at all. It would simply do some calculations, but never write its result back to the `output` array. However, due to the fact, that the `output` array does not get reset between consecutive calls, the unit tests would still run through without indicating an error (they still had the correct values from the previous function invocation). This can of course potentially apply to all unit test cases where more than one function is called and expected to produce the same `output`. 

A simple and straightforward fix to this issue would be to overwrite the `output` between function invocations. This can be done by adding a single line `act[i] = 0;` in the `validate()` function [here](https://github.com/ARM-software/CMSIS_5/blob/0a42364521c589690ffc5c63bdf662c05bd0f27a/CMSIS/NN/Tests/UnitTest/TestCases/Utils/validate.h#L38) (for `validate_s16()` analogously). 

To nitpick some more: In theory, the same can be said for the intermediate buffer `ctx.buf`. It, or rather its underlying memory, should (ideally) also not be left untouched. While it does get `free`ed and `malloc`ed between consecutive calls, the chances of `malloc` returning the same memory (buffer) are quite high. Thus, in the (admittedly very rare) case that only a single `im2col` iteration would be needed to compute the `output` tensor, a similarly implemented followup function could have a broken `im2col` implementation and "cheat" its way through the unit tests by simply relying on the successful computation of the `im2col` buffer from the previous function. 

What both these cases come down to is that one should ideally reset all output (and ideally also intermediate) buffers between consecutive unit tests to either zero (or even better random) values. This ensures that functions do not "cheat" and that they have to compute all needed values themselves.

felix-johnny:
Clear the output buffers within a test case if the same
buffer is reused.

Fixes #1532

### DIFF
--- a/CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_fast_s16/test_arm_convolve_fast_s16.c
+++ b/CMSIS/NN/Tests/UnitTest/TestCases/test_arm_convolve_fast_s16/test_arm_convolve_fast_s16.c
@@ -91,6 +91,7 @@ void int16xint8_arm_convolve_fast_s16(void)
 
     TEST_ASSERT_EQUAL(ARM_CMSIS_NN_SUCCESS, result);
     TEST_ASSERT_TRUE(validate_s16(output, output_ref, output_ref_size));
+    memset(output, 0, sizeof(output));
 
     buf_size = arm_convolve_fast_s16_get_buffer_size(&input_dims, &filter_dims);
     ctx.buf = malloc(buf_size);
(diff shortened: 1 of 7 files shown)


grasci-arm:

mansnils (approved):

PR #1550, opened 2022-08-15
CMSIS-NN: Re-deliver block process for DW conv

felix-johnny:
A force push error in PR
https://github.com/ARM-software/CMSIS_5/pull/1545
meant that the actual change got overwritten. This
is a re-delivery of the change.

DW conv for MVE is changed to process in smaller
blocks. For most cases this results in lower scratch
buffer requirement.


grasci-arm:

PR #1574, opened 2022-10-27
Added macros with bit definitions for Pos/Msk for PTIM peripheral acess structure

Masmiseim36:
Added macros with bit definitions for Pos/Msk for PTIM peripheral acess structure for Cortex-A


grasci-arm:

PR #1575, opened 2022-10-30
Make the IRQ Table extern definable.

Masmiseim36:
This makes the IRQ-table extern definable. With this it is possible to define Interrupt Service Routines in the CMSIS way with week defined IRQ-Routines which can be overridden in the application code.

Example:
// …
\_\_attribute__ ((weak, alias("InteruptDefaultHandler"))) void CoreTimer_IRQHandler (void);
// …

IRQHandler_t IRQTable [IRQ_GIC_LINE_COUNT] =
{
// …
CoreTimer_IRQHandler
// …
};

In Addition a default IRQ-Handler is implemented which calls the Interrupt-Service-Routine.

### DIFF
--- a/CMSIS/Core_A/Source/irq_ctrl_gic.c
+++ b/CMSIS/Core_A/Source/irq_ctrl_gic.c
@@ -1,11 +1,11 @@
 /**************************************************************************//**
  * @file     irq_ctrl_gic.c
  * @brief    Interrupt controller handling implementation for GIC
- * @version  V1.1.1
- * @date     29. March 2021
+ * @version  V1.2.0
+ * @date     30. October 2022
  ******************************************************************************/
 /*
- * Copyright (c) 2017-2021 ARM Limited. All rights reserved.
+ * Copyright (c) 2017-2022 ARM Limited. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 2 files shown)


grasci-arm:

CMSIS/Core_A/Source/irq_ctrl_gic.c
```diff
@@ -65,6 +71,14 @@ __WEAK int32_t IRQ_SetHandler (IRQn_ID_t irqn, IRQHandler_t handler) {
   return (status);
 }
 
+/// The Interrupt Handler.
+__WEAK void IRQ_Handler (void) {
+  IRQn_Type irqn = GIC_AcknowledgePending ();
+  if (irqn < (IRQn_Type)IRQ_GIC_LINE_COUNT)
+    IrqTable[irqn]();
```
JonatanAntoni: Please add braces `{ }` around the block to prevent merge anomalies.

```suggestion
  if (irqn < (IRQn_Type)IRQ_GIC_LINE_COUNT) {
    IrqTable[irqn]();
  }
```

JonatanAntoni (changes requested):

CMSIS/Core_A/Source/irq_ctrl_gic.c
Masmiseim36: I’ve added braces around the statement.

CMSIS/Core_A/Source/irq_ctrl_gic.c
```diff
@@ -1,11 +1,11 @@
 /**************************************************************************//**
  * @file     irq_ctrl_gic.c
  * @brief    Interrupt controller handling implementation for GIC
- * @version  V1.1.1
- * @date     29. March 2021
+ * @version  V1.1.2
```
JonatanAntoni: Shall we go for a minor version increment in this case?
I think we are adding a feature not just fixing a bug, right?
```suggestion
 * @version  V1.2.0
```

JonatanAntoni (commented):
I recognized we need to update the "component version" in `Cversion` here as well:
https://github.com/ARM-software/CMSIS_5/blob/2d388f6bac996ca4254e9544d9fe0704a4ad45e1/ARM.CMSIS.pdsc#L2210

We need to decide if this is a minor or a patch update.

CMSIS/Core_A/Source/irq_ctrl_gic.c
Masmiseim36: I’ve updated the version number as suggested.

PR #1578, opened 2022-11-01
Replacing secrets to plain text

spcaipers-arm:
This change allows fork-based PR to trigger Actions without failing due to the unavailability of secrets.


spcaipers-arm:
Hi @JonatanAntoni, the changes are working in my fork:
https://github.com/spcaipers-arm/CMSIS_5/actions/runs/3368774105

The changes will only have effect once this PR is merged.
Unfortunately, the Arm-Software org does not allow to use of repo environments, so we have to hardcode the AWS data in the workflow. 

Also, please have a look at the tests. There are some tests failing which I am not sure is expected.

FYI: @jkrech

JonatanAntoni (approved):

Issue #1580, opened 2022-11-04
Cortex-A build fix


ua1arn:
![image](https://user-images.githubusercontent.com/34999007/199962909-4944a999-f4e4-4032-b035-1e13cce6830e.png)

Case mismatch.

JonatanAntoni:
Thanks for raising this. I hope @Masmiseim36 can provide a fix shortly.

Masmiseim36:
Thanks for the hint. I should have noticed that one. Fix is here:
https://github.com/ARM-software/CMSIS_5/pull/1581

PR #1581, opened 2022-11-04
Fix for Issue #1580

Masmiseim36:
Fix for Issue #1580


grasci-arm:

PR #1584, opened 2022-11-10
Two-tier approach to overcome GitHub permission restriction

spcaipers-arm:
Using a two-tier approach to overcome GitHub permission restriction on pull requested-based runs.

The caller-corevalidation.yml is triggered by PRs (with limited permission if forked), it collects PR number to be passed to the called corevalidation.yml (this run has  full permissions, e.g. to assume-role and consume secrets).

The corevalidation workflow will feedback on its status to the PR with a "CoreValidation" check name.

Caveat: The corevalidation.yml code is, by GH design, run from the base branch, not from the PR. So, changes on integration-test.yml file only take effect when merged to the base branch (e.g. main)

### DIFF
--- a/.github/workflows/caller-corevalidation.yml
+++ b/.github/workflows/caller-corevalidation.yml
@@ -1,0 +1,27 @@
+name: Caller CoreValidation
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    paths:
+      - .github/workflows/caller-corevalidation.yml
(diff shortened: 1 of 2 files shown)


spcaipers-arm:
Hi @JonatanAntoni ,

A similar solution you saw for avhclient.
I tested it on my fork, looks alright --> https://github.com/spcaipers-arm/CMSIS_5/pull/4

Again, it only works after merging this PR.

PR #1587, opened 2022-11-14
Harmonize compiler settings for Cortex-M and Cortex-A

Masmiseim36:
Both versions (Cortex-M and Cortex-A) are better comparable now.
This brings no functional change

The compiler specific abstractions seem to have been copied from the Core_M implementation to the Core_A implementation sometime in the past and developed apart. With this pull request I tried to aligned them again. But I think it would be best to have only one generic implementation to avoid code duplication. Core specific implementations could still be in separate files, but the most part of it could be generic.
What do you think, is it worth to work in this direction? Why has it been copied in the first place?

### DIFF
--- a/CMSIS/Core_A/Include/cmsis_armcc.h
+++ b/CMSIS/Core_A/Include/cmsis_armcc.h
@@ -1,8 +1,8 @@
 /**************************************************************************//**
  * @file     cmsis_armcc.h
- * @brief    CMSIS compiler specific macros, functions, instructions
- * @version  V1.0.5
- * @date     05. May 2021
+ * @brief    CMSIS compiler ARMCC (Arm Compiler 5) header file
+ * @version  V1.0.6
+ * @date     13. November 2022
  ******************************************************************************/
 /*
  * Copyright (c) 2009-2021 Arm Limited. All rights reserved.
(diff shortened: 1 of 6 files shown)


grasci-arm:

JonatanAntoni (approved):
LGTM. Reducing duplication would be a good step forward of course. But we need to be careful not to break existing stuff. Would you be able to come up with a file/dir layout that would work for you?

ghost (approved):
Changes look fine to me.

ua1arn:
Bay the way, this PR also solve some errors then use riscv-none-elf-gcc for build CMSIS-DSP

ua1arn:
Only one ifndef pair inside cmsis_gcc.h solve error when use riscv-none-elf-gcc for compile CMSIS-DSP...
Or this offtopic in organization?

JonatanAntoni:
@ua1arn, well Risc-V of course is not Arm-business ;)
But if you want to raise the required change as PR ...

ua1arn:
Well. Similar local modifications used then compile for x86 targets...

JonatanAntoni:
Sure. x86 is raised repeatedly as well. I am happy to receive contributions in these areas.

ua1arn:
Main direction - isolate inlines like __QASX __SHADD16 in dsp-related headers. Machine-dependencies in CMSIS-DSP more complete.

ua1arn:
Tray this:
image

PR #1592, opened 2022-12-03
CMSIS-DAP: DAP_XXX_Transfer() counts too much in case of break in the first loop

rgrr:
Bug is, that parsing in the DAP_XXX_Transfer() functions overflow the input stream if there is a "DAP_TransferAbort".  In this case "request" is incremented, but "request_count" is not decremented as it should be.

The fix solves this by using another loop construct which always decrements "request_count".

This bug is present in all versions of CMSIS!

This is a modified version of https://github.com/ARM-software/CMSIS_5/pull/1503 which also takes DAP_JTAG_Transfer() into account.


grasci-arm:

RobertRostohar:
Hi @rgrr,

thank you for analyzing this issue and providing a solution which seems to solve the problem.

PR #1596, opened 2022-12-17
Address unsupported nested external declaration #617

ya-mouse:
This should address an issue #617 for Clang compilers.


grasci-arm:

ya-mouse:
This change is mutually exclusive with #1597

JonatanAntoni:
Hi @ya-mouse,

Thanks for raising this. I tend to prefer your second solution in order not to duplicate all the code. GCC and Clang seem to use the same code so we should not duplicate the code.

Cheers,
Jonatan

ya-mouse:
Ok, proceeding with another PR then.

PR #1597, opened 2022-12-17
Address unsupported nested external declaration #617

ya-mouse:
This should address an issue #617 for Clang compilers.

This change is mutually exclusive with #1596

### DIFF
--- a/CMSIS/Core/Include/cmsis_gcc.h
+++ b/CMSIS/Core/Include/cmsis_gcc.h
@@ -1,8 +1,8 @@
 /**************************************************************************//**
  * @file     cmsis_gcc.h
  * @brief    CMSIS compiler GCC header file
- * @version  V5.4.1
- * @date     27. May 2021
+ * @version  V5.4.2
+ * @date     17. December 2022
  ******************************************************************************/
 /*
  * Copyright (c) 2009-2021 Arm Limited. All rights reserved.
@@ -132,13 +132,13 @@ __STATIC_FORCEINLINE __NO_RETURN void __cmsis_start(void)
 {
   extern void _start(void) __NO_RETURN;
 
-  typedef struct {
+  typedef struct __copy_table {
     uint32_t const* src;
     uint32_t* dest;
     uint32_t  wlen;
   } __copy_table_t;
 
-  typedef struct {
+  typedef struct __zero_table {
     uint32_t* dest;
     uint32_t  wlen;
   } __zero_table_t;


grasci-arm:

JonatanAntoni:
Hi @ya-mouse,

Thanks for raising this.

It looks like there are no real differences between GCC and Clang, expect a small detail in `__cmsis_start`. Why do we need this? The location of the type definitions has been discussed multiple times, already. If we decide to move them out of the function  and pollute the global namespace, this change could be done for both, GCC and Clang.

So far, I think we should keep the type defines as local as possible. As these types are never used outside of the `__cmsis_start` function there is absolutely no need to define the types in global scope. 

Cheers,
Jonatan

ya-mouse:
@JonatanAntoni this change addresses the issue #617 that is blocking e.g. TinyGo CMSIS version bump https://github.com/tinygo-org/tinygo/pull/3339 while it use Clang to compile.
I can drop the `typedef` and keep it just as structure name to not pollute with new types. I can do it just in `cmsis_gcc.h`, dropping clang case.

JonatanAntoni:
In #617 we decided not to "fix" this issue, at least not by moving the typedefs to global scope. The issue is just a Compiler warning which is correct in general but doesn't apply in this specific case. Hence, we might add pragmas to disable the warning in this place.

Another aspect of your PR is adding upstream Clang support, which is fine for me. If Clang is basically happy with the `compiler_gcc.h` and doesn't need changes, we can include this file in `cmsis_compiler.h` for both GCC and Clang. We just need to be careful that we still include the `compiler_armclang.h` for Arm Compiler which is based on Clang and hence has the same predefined macros set.

ya-mouse:
Actually, there is much easier fix. Just need to name the typedef'ed structure, two lines changes. And it fixes the problem.
The hint is in error itself: `error: unsupported: anonymous type given name for linkage purposes by typedef declaration after its linkage was computed; ***add a tag name here*** to establish linkage prior to definition`.
Going to update the PR.

JonatanAntoni:
Looks good to me. Could you please update the file header (version and date) as indicated by the check?

ya-mouse:
Sure. Thanks!

ya-mouse:
@JonatanAntoni should be good for tests run & final review

ya-mouse:
@JonatanAntoni I'm a bit confused with force pushed change as disconnnected commit instead of the develop branch like for other changes: https://github.com/ARM-software/CMSIS_5/pulls?q=is%3Apr+is%3Aclosed

JonatanAntoni:
@ya-mouse, sorry, I clicked the "update branch" button which automatically rebased your branch.
Do you intend to add further commits? If so, please feel free to force-push again. If not, the branch looks good to be merged.

ghost (approved):
Change looks good.

ya-mouse:
@JonatanAntoni no, I don't plan to add more commits. Thanks!

PR #1600, opened 2022-12-19
Core A: cache: instr mva

yroeht:
As documented by ARM at:
    https://developer.arm.com/documentation/ddi0301/h/Babhejba

Being able to do all these options would be useful for Zephyr, which already provides interfaces for these.

I wasn't sure what the C in xMVAC stands for, so I chose this name, I hope it is correct.


grasci-arm:

ghost (approved):
Looks correct to me.

PR #1611, opened 2023-03-17
Core(M): Fix endless loop issue with non-optimized IAR builds

TTornblom:
This is an IAR fix for the problem described in
https://github.com/ARM-software/CMSIS_5/issues/620

IAR builds can not align the stack to the cache line size and thus the invalidation is done in separate steps for the three variables.

Fix validated on STM32H7 HW.


grasci-arm:

JonatanAntoni:
@TTornblom, please could you update the file version of the changed files.

ghost (approved):

PR #1614, opened 2023-03-24
Enable support for TI Arm Clang Compiler (tiarmclang)

evodius96:
- Add new cmsis_tiarmclang.h and modified existing Core header files
- Add linker command files (linker scripts) to supported CoreValidation targets
- Adjust CoreValidation to avoid compiler removal of empty "delay loops"

### DIFF
--- a/CMSIS/Core/Include/cmsis_tiarmclang.h
+++ b/CMSIS/Core/Include/cmsis_tiarmclang.h
@@ -1,0 +1,1510 @@
+/**************************************************************************//**
+ * @file     cmsis_tiarmclang.h
+ * @brief    CMSIS compiler tiarmclang header file
+ * @version  V5.6.0
+ * @date     20. March 2023
+ ******************************************************************************/
+/*
+ * Copyright (c) 2009-2023 Arm Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the License); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*lint -esym(9058, IRQn)*/ /* disable MISRA 2012 Rule 2.4 for IRQn */
+
+#ifndef __CMSIS_TIARMCLANG_H
+#define __CMSIS_TIARMCLANG_H
+
+#pragma clang system_header   /* treat file as system include file */
+
+/* CMSIS compiler specific defines */
+#ifndef   __ASM
+  #define __ASM                                  __asm
+#endif
+#ifndef   __INLINE
+  #define __INLINE                               __inline
+#endif
+#ifndef   __STATIC_INLINE
+  #define __STATIC_INLINE                        static __inline
+#endif
+#ifndef   __STATIC_FORCEINLINE
+  #define __STATIC_FORCEINLINE                   __attribute__((always_inline)) static __inline
+#endif
+#ifndef   __NO_RETURN
+  #define __NO_RETURN                            __attribute__((__noreturn__))
+#endif
+#ifndef   __USED
+  #define __USED                                 __attribute__((used))
+#endif
+#ifndef   __WEAK
+  #define __WEAK                                 __attribute__((weak))
+#endif
+#ifndef   __PACKED
+  #define __PACKED                               __attribute__((packed, aligned(1)))
+#endif
+#ifndef   __PACKED_STRUCT
+  #define __PACKED_STRUCT                        struct __attribute__((packed, aligned(1)))
+#endif
+#ifndef   __PACKED_UNION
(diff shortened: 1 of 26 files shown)


grasci-arm:

CMSIS/Core/Include/cmsis_tiarmclang.h
```diff
@@ -0,0 +1,1510 @@
+/**************************************************************************//**
+ * @file     cmsis_tiarmclang.h
+ * @brief    CMSIS compiler tiarmclang header file
+ * @version  V5.6.0
```
JonatanAntoni: This is a new file and version should start at 1.0.0

CMSIS/Core/Include/core_armv81mml.h
```diff
@@ -97,6 +97,29 @@
     #define __DSP_USED         0U
   #endif
 
+#elif defined (__ti__)
```
JonatanAntoni, line 100: Adding this for every compiler seems not to be maintainable on the long run when adding more compilers. We need to look for another solution. Can we move this somehow into the compiler header? Or reduce to be compiler independent by adding a define-mapping to compiler headers?

CMSIS/CoreValidation/Source/CV_CoreFunc.c
```diff
@@ -32,6 +32,12 @@ static void TC_CoreFunc_IPSR_IRQHandler(void) {
   irqXPSR = __get_xPSR();
 }
 
+#if defined(__ti__)
```
JonatanAntoni: This is only test code, so performance is not big issue.
Hence, I'd add `__NOP` in any case and remove compiler dependency here.

CMSIS/CoreValidation/Source/CV_CoreFunc.c
```diff
@@ -307,6 +313,9 @@ void TC_CoreFunc_IPSR (void) {
 #if defined(__CC_ARM)
 #define SUBS(Rd, Rm, Rn) __ASM volatile("SUBS " # Rd ", " # Rm ", " # Rn)
 #define ADDS(Rd, Rm, Rn) __ASM volatile("ADDS " # Rd ", " # Rm ", " # Rn)
+#elif defined(__ti__)
```
JonatanAntoni: Why is this duplication of the else-tree required?
Is it because `__GNUC__` is defined?

JonatanAntoni: Perhaps we need to move these defines into the compiler header?

JonatanAntoni (changes requested):
Please consider updating all touched files versions and dates

ghost (changes requested):
File version numbers and File dates need to be checked and corrected.

CMSIS/CoreValidation/Source/CV_CoreFunc.c
evodius96: OK will do

CMSIS/CoreValidation/Source/CV_CoreFunc.c
evodius96: Yes, `__GNUC__` is also defined.  However, I can just add a` !defined(__ti__)` to the `GNUC `clause and get the same behavior without duplicating the macros.  it's worth exploring moving these defines into the compiler headers.

CMSIS/Core/Include/cmsis_tiarmclang.h
evodius96: No problem.   I'm assuming, then, that the file version and date reflect the last file modification?  In what way does the version increment when the file is changed?

CMSIS/Core/Include/core_armv81mml.h
evodius96: In general, I agree -- there is a lot of duplication here, and there are also some target specific dependencies in a handful of the files (depending on whether the target cares about the compiler enabling FPU instructions, DSP instructions, or not).  Also, tiarmclang also defines __GNUC__, which ties it together.  I'm willing to do what I can for tiarmclang, but I think someone needs to look at the design holistically and then refactor in a way that makes sense.  I think that's beyond the scope of this pull request, although I absolutely understand why you'd call it out.

CMSIS/Core/Include/cmsis_tiarmclang.h
JonatanAntoni: Yes, reason behind having individual file versions and dates is that still people are used to ship copies of cmsis files (instead of relying on official releases of the pack). In order to sort out user feedback we need to specifically know which version if a files a user received. The version shall follow [SemVer](https://semver.org/lang/de/) rules, i.e., increment major part on non-backward compatible changes (should be avoided), minor part for new features, and patch part for bug-fixes only.

CMSIS/Core/Include/core_armv81mml.h
JonatanAntoni: Yes, agree. We should not do to much at once. I think we eventually can move the defines into the compiler headers. Let's leave it as is for now. I'll put me a note to address this change in CMSIS 6.

ghost (approved):

CMSIS/Core/Include/cmsis_tiarmclang.h
JonatanAntoni: Do you plan to set this to 1.0.0 for the initially contributed version?

evodius96: My second commit makes this change to V1.0.0 -- is that not visible?  If necessary, I can squash the commits.  We use a different review infrastructure, so I apologize if I am not using pull requests optimally.

PR #1617, opened 2023-04-10
Adding MPU macros and improving docs

M-Plichta:
Functional changes:
- There already were macros for MPU_RBAR shareability field (`ARM_MPU_SH_NON`, `ARM_MPU_SH_OUTER` and `ARM_MPU_SH_INNER`), so I added macros for the access permissions and execute-never fields.
- Added some handy functions and macros for better readability, like `ARM_MPU_TYPE` function and the macro `MAIR_ATTR`.
- Added input macros for `ARM_MPU_ATTR`

Docs:
- Changed wording for clarity and consistency with the Architecture Reference Manual.

PR #1619, opened 2023-04-16
Updating docs

M-Plichta:
Updated docs and examples based on my [previous pull request](https://github.com/ARM-software/CMSIS_5/pull/1617).

Unsure about the change to `ARM_MPU_RBAR` on line 148. The other macros did not have their implementations so `ARM_MPU_RBAR` seemed out of place. Happy to change back if I'm wrong.


grasci-arm:

PR #1622, opened 2023-04-19
Correct bit field width in EWIC Event Number ID Reg and TCM Error Bank Reg

joseph-yiu:
EWIC_NUMID: NUMEVENT is 16-bit in the TRM.
TEBR0/TEBR1: BANK filed is 3-bit in the TRM.
Revision updated.


ghost (approved):
The changes are correct.

PR #1624, opened 2023-05-01
IRQn_Type: Ensure enum size matches maximum value.

axel-simon:
This change fixes a bug where a compiler choses a signed byte to store the IRQn_Type enum type as all listed enum values fit a signed byte. This causes undefined behavior when calling functions such as NVIC_EnableIRQ with interrupt channels greater than 127 as these do not fit a signed byte. As a result, interrupt channels higher than 127 cannot be enabled, disabled, or triggered.

Fix this by listing the highest possible interrupt channel in each enum declaration which ensures that a storage type is chosen that can accommodate all valid interrupt channels.

Tested on a MEC1527 devices by triggering interrupt channel 171 (watchdog) and observing that the ISR is executed.

### DIFF
--- a/Device/ARM/ARMCM0/Include/ARMCM0.h
+++ b/Device/ARM/ARMCM0/Include/ARMCM0.h
@@ -48,16 +48,8 @@ typedef enum IRQn
 
 /* -------------------  Processor Interrupt Numbers  ------------------------------ */
   Interrupt0_IRQn               =   0,
-  Interrupt1_IRQn               =   1,
-  Interrupt2_IRQn               =   2,
-  Interrupt3_IRQn               =   3,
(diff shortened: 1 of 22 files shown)


grasci-arm:

JonatanAntoni (changes requested):
Can we keep the enum members `Interrupt1_IRQn` to `Interrupt9_IRQn`, please?
They might be used by test applications. Removing existing declarations would be a breaking change we should avoid.

Please take care about the file headers and update the versions (increment patch level) and dates.

JonatanAntoni (approved):

axel-simon:
Thank you for the quick turnaround!

PR #1627, opened 2023-05-22
Core/Include: Fix GCC __USAT() macro function + Improve USAT doc

benjarobin:
First argument of USAT instruction is a signed number, not an unsigned value. Update documentation everywhere else to clarify that.


JonatanAntoni:
Hi @benjarobin, thanks for pointing this out. We are in the process of moving to CMSIS 6. Once the new major repo is public, we need to address it there.

PR #1630, opened 2023-05-30
Added Virtual Timer support

Mayankbht:
Update core_ca.h for adding Virtual Timer support to CMSIS_5


grasci-arm:

PR #1633, opened 2023-06-02
Enhance the cast from uint32_t to uint32_t* by intermediate casting through uintptr_t in __NVIC_SetVector/GetVector

M-Bab:
This allows the header to be compiled on a non-32bit architecture without the warning int-to-pointer-cast. It is also a clean approach when turning a uint32_t value to a pointer and should not add any additional cycles on a 32bit architecture.

The next obvious question would be: Why should I ever compile these very specific headers on a different architecture?

The answer: For example for _unit testing the embedded code on an x64 host_. When trying to do this, these files alone emits dozens of warnings:

```
... Include/core_cm0plus.h: In function ‘__NVIC_SetVector’:
... Include/core_cm0plus.h:890:23: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
  890 |   uint32_t *vectors = (uint32_t *) ((uint32_t) SCB->VTOR);
      |                       ^
... Include/core_cm0plus.h: In function ‘__NVIC_GetVector’:
... Include/core_cm0plus.h:909:23: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
  909 |   uint32_t *vectors = (uint32_t *) ((uint32_t) SCB->VTOR);
```

Depending on the Werror policy during unit-testing it might stop your efforts immediately.

### DIFF
--- a/CMSIS/Core/Include/core_armv81mml.h
+++ b/CMSIS/Core/Include/core_armv81mml.h
@@ -3606,7 +3606,7 @@ __STATIC_INLINE void NVIC_DecodePriority (uint32_t Priority, uint32_t PriorityGr
  */
 __STATIC_INLINE void __NVIC_SetVector(IRQn_Type IRQn, uint32_t vector)
 {
-  uint32_t *vectors = (uint32_t *)SCB->VTOR;
+  uint32_t *vectors = (uint32_t *) ((uintptr_t) SCB->VTOR);
   vectors[(int32_t)IRQn + NVIC_USER_IRQ_OFFSET] = vector;
   __DSB();
 }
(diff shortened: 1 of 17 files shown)


grasci-arm:

JonatanAntoni:
Thanks for raising this. Not sure, if for unit testing higher level application code should make use of another hardware abstraction. But I am basically fine with "silencing" the warnings. As we are in the process of moving CMSIS to the next level, would you mind raising this PR at https://github.com/ARM-software/CMSIS_6?

M-Bab:
> As we are in the process of moving CMSIS to the next level, would you mind raising this PR at https://github.com/ARM-software/CMSIS_6?

Done: https://github.com/ARM-software/CMSIS_6/pull/10

M-Bab:
Closed because it was merged in CMSIS 6.

PR #1644, opened 2023-08-08
CMSIS: add ARMv8-A AARCH64 support

JiafeiPan:
Currently RTOS can run Cortex-A AARCH64 platforms, so add ARMv8-A AARCH64 support in CMSIS, it includes Core header file support for Cortex-A53 and Cortex-A55, cache driver, MMU driver and GICv3 driver.

This patches are merged from a patch set with the contributors list in the following Signed-off-by list.


JonatanAntoni:
Hi @JiafeiPan, thanks for your contribution. CMSIS 5 is in maintenance only and not going to be enhanced anymore. New development takes place in https://github.com/ARM-software/CMSIS_6. 

Please be aware that we are currently restructuring CMSIS-Core, see ARM-software/CMSIS_6#38. Please consider contributing this one based on the restructuring. We are going to have an a-profile folder for all files specific for Cortex-A. I think Armv8-A 64-bit can go there. Lets check that we only add AArch64-specific stuff and reuse all the generic stuff from existing files.

PR #1658, opened 2024-01-18
added bss start and end to zero.table

amir-skn:
The missing lines were resulting in the BSS section not being properly initialized, which might lead to undefined behavior in the program as uninitialized global and static variables may contain garbage values.

### DIFF
--- a/Device/ARM/ARMCM4/Source/GCC/gcc_arm.ld
+++ b/Device/ARM/ARMCM4/Source/GCC/gcc_arm.ld
@@ -127,7 +127,7 @@ SECTIONS
   /*
    * SG veneers:
    * All SG veneers are placed in the special output section .gnu.sgstubs. Its start address
-   * must be set, either with the command line option ‘--section-start’ or in a linker script,
+   * must be set, either with the command line option â€˜--section-startâ€™ or in a linker script,
    * to indicate where to place these veneers in memory.
    */
 /*
@@ -170,6 +170,10 @@ SECTIONS
   {
     . = ALIGN(4);
     __zero_table_start__ = .;
+
+    LONG (__bss_start__)
+    LONG ((__bss_end__ - __bss_start__) / 4)
+
     /* Add each additional bss section here */
 /*
     LONG (__bss2_start__)


JonatanAntoni:
Hi @amir-skn,

May I ask you which C library you are using?
By default, we assume newlib-nano being used. In this case `bss` section should be initialized implicitly by the library pre-main. If you have another pre-main or run completely without any default startup you need to take care yourself.

amir-skn:
Hi @JonatanAntoni 
Thank you for your response. I'm using newlib-nano. However, the linker couldn't locate the `_start` symbol within the startup assembly code. To address this, I defined a `_start` function myself, which jumps to the main function and then calls `_exit`. Based on your explanation, I realize that my approach might be incorrect. May I ask how can I get the original _start function working?

JonatanAntoni:
Hi @amir-skn,

I see, this depends on the library spec you are using. I guess you are using the `nosys.specs` which has no `_start` or `_end` symbols. In this case you need to provide the implementation on your own and need to adopt the linker script, accordingly.

You could try to use `nano.specs` instead.

amir-skn:
Thank you.

amir-skn:
I checked the definition of _start (or __PROGRAM_START which points to _start) in case of using gcc:
https://github.com/ARM-software/CMSIS_5/blob/b9667965e0c41d7ee0c63983beccc9b8cd6b0f7f/CMSIS/Core/Include/cmsis_gcc.h#L131-L164
 It uses `__zero_table_start__` and `__zero_table_end__` to zero-initiliase the secition. Which suggests that the bss section should be included in `.zero.table` to be a part of zero initialisation. The comment also suggests this for other optional extra bss sections but doesn't include the original one.

 https://github.com/ARM-software/CMSIS_5/blob/a75f01746df18bb5b929dfb8dc6c9407fac3a0f3/Device/ARM/ARMCM4/Source/GCC/gcc_arm.ld#L169-L179


On top of that the amended Linker Script in the pull request comes with startup assembly file in the same directory which includes a section to zero initialise the bss section and again it uses  `__zero_table_start__` and `__zero_table_end__`.

https://github.com/ARM-software/CMSIS_5/blob/a75f01746df18bb5b929dfb8dc6c9407fac3a0f3/Device/ARM/ARMCM4/Source/GCC/startup_ARMCM4.S#L100-L103

JonatanAntoni:
Hi @amir-skn,

Please carefully check your C library startup. The newlib CRT startup (`_start`) is expected to zero-initialize the bss section. So there should be no need to initialize bss section explicitly in `_cmsis_start`. Hence, we removed bss section from zero table as an optimization.

If you are using a different library configuration which does not zero-initialize bss section as part of `_start` or not having a CRT startup at all, you can still add bss into zero table by patching your projects linker script.

However, from CMSIS point of view we assume newlib-nano with CRT startup is used. Hence, we don't include bss section in zero table by default.

amir-skn:
Thank you. Resolved.

PR #1664, opened 2024-03-23
Fix typos in docs

RoboSchmied:
fix typos


JonatanAntoni:
Hi @RoboSchmied,

Thanks for pointing this typos out. Unfortunately, renaming defines should be considered a braking change. We shouldn't do this outside of a major version increment.

We could easily update the documentation text of course. But for the defines we'd need to keep the old defines (with the typos) for backward compatibility reason; e.g., marked as deprecated.

Would you want to update your PR, accordingly? Perhaps, you want to contribute a similar change to the next major version of CMSIS here: https://github.com/ARM-software/CMSIS_6. Unfortunately, as CMSIS 6.0.0 has been released already, we need to keep the typos in the defines for backward compatibility reasons there as well. But we can add the correctly typed versions in addition.

Thanks again,
Jonatan

RoboSchmied:
Hello @JonatanAntoni 

Thank you for your review and explanation.

> We could easily update the documentation text of course. But for the defines we'd need to keep the old defines (with the typos) for backward compatibility reason; e.g., marked as deprecated.
> 
> Would you want to update your PR, accordingly? 

Yes, sure. Done.

> Thanks again, Jonatan0

I thank you, Sir.

JonatanAntoni (approved):

PR #1674, opened 2024-07-30
Fix typos in documentation

Aric5301:
I also created a PR for CMSIS-RTX:  https://github.com/ARM-software/CMSIS-RTX/pull/44


JonatanAntoni:
Thanks @Aric5301 for raising this. Would you check the same in recent https://github.com/ARM-software/CMSIS_6?

JonatanAntoni (approved):

Aric5301:
> Thanks @Aric5301 for raising this. Would you check the same in recent https://github.com/ARM-software/CMSIS_6?

It doesn't exist in CMSIS-6 because of the split between RTX and CMSIS.
Thank you for the quick response Jonatan!

PR #1687, opened 2025-06-25
Update cmsis_armcc.h

NVMarkovich:
doxygen ignores /*@} for end of group. It needs /** @} instead.


JonatanAntoni:
Not that CMSIS_5 is not actively maintained anymore.
You may want to check https://github.com/ARM-software/CMSIS_6

